PR #82, opened 2021-08-25
Select skynet-js version automatically

firyx:


MSevey (approved):

ro-tex (approved):
LGTM.

Printing all `skynet-js` versions twice might be a bit noisy. Maybe only do it when we return an error? Not important enough to block the PR, though.

firyx:
> LGTM.
> 
> Printing all `skynet-js` versions twice might be a bit noisy. Maybe only do it when we return an error? Not important enough to block the PR, though.

in the happy path the log prints only the selected version,
python output from `module: ansible.builtin.shell` is printed only if no version is found

PR #88, opened 2021-08-30
Fix UFW firewall setup

firyx:


ro-tex (approved):

playbooks/tasks/portal-firewall-ufw-setup.yml
```diff
@@ -0,0 +1,61 @@
...
+      community.general.ufw:
+        default: allow
+        direction: outgoing
+
+    - name: Deny incoming traffic by default
+      community.general.ufw:
+        default: deny
+        direction: incoming
+
+    - name: Allow specific incoming ports
```
MSevey, line 27: Follow Up: We should add comment explaining why each port is needed for reference in the future.

MSevey (approved):

PR #89, opened 2021-08-30
Fix docker system prune

firyx:


playbooks/tasks/portal-docker-services-start.yml
```diff
@@ -74,7 +80,9 @@
 
 # Prune docker
 - name: Prune docker
-  command: docker system prune
+  ansible.builtin.command: docker system prune --force
```
MSevey, line 83: If we are pruning before we start the services, is this doing anything?

playbooks/tasks/xxx-move-to-portal-role.yml
```diff
@@ -87,8 +87,11 @@
   when: create_env_file_result.changed or sia_container_info_result.container == None or not sia_container_info_result.container.State.Running
 
 # xxx use community.docker.docker_prune (requires community.docker upgrade) with long timeout
+# Prune docker
 - name: Prune docker
-  command: docker system prune --force
+  ansible.builtin.command: docker system prune --force
```
MSevey, line 92: shouldn't this be before the sia docker container start?

playbooks/tasks/portal-docker-services-start.yml
firyx: If no docker container/image/network was run/pulled/built/set on this machine, it's doing nothing,
otherwise it's pruning things from previous docker executions (e.g. from previous deployments).

playbooks/tasks/xxx-move-to-portal-role.yml
firyx: Added docker prune also before starting services.

MSevey (approved):

ChrisSchinnerl (approved):

PR #91, opened 2021-08-31
Add xxx* files to gitignore

firyx:


playbooks/tasks/portal-role-task.yml
```diff
@@ -1,5 +1,7 @@
 ---
 
+# xxx This ansible playbook task, should be moved to portal role
+
```
ro-tex, line 4: Shouldn't this comment be changed now?

.gitignore
```diff
@@ -13,4 +13,7 @@ ansible_collections
 inventory/*
 my-logs/*
 my-vars/*
-roles
\ No newline at end of file
+roles
+
+# Files in development with filename starting 'xxx'
+xxx*
```
kwypchlo, line 19:
```suggestion
xxx*

```

playbooks/tasks/portal-role-task.yml
firyx: I just added this comment now.

This file is now an ansible playbook task (not an ansible role task).

Playbook is like a script,
role is a reusable library used by playbooks.

Once it's finished this file (task) should be moved to separate ansible role repo (https://github.com/SkynetLabs/ansible-role-skynet-webportal), where it should belong so Skynet playbooks and community playbooks can execute skynet-webportal role.

It's not yet now part of the role, because during development I would need to update the role task, reinstall the role and run/debug the playbook each time.

kwypchlo (approved):

ro-tex (approved):

PR #95, opened 2021-09-02
Add check for NGINX starting up and common time vars

MSevey:
This PR adds the check that the NGINX docker container is up and not restarting.  It has a 15min timeout.

I added a note saying that the check is temporary, needed for next week's deploy. But it should be an instant pass in normal cases so there is no real harm in leaving it in.

### DIFF
--- a/playbooks/tasks/portals-check-nginx.yml
+++ b/playbooks/tasks/portals-check-nginx.yml
@@ -1,0 +1,7 @@
+# Wait until nginx is no longer restarting
+- name: Check for Restarting in docker ps
+  shell: docker ps | grep nginx
+  register: docker_nginx
+  until: docker_nginx.stdout.find("Restarting") == -1
+  delay: 1
+  retries: "{{ fifteen_minutes }}"
--- a/playbooks/tasks/portals-check-nginx.yml
+++ b/playbooks/tasks/portals-check-nginx.yml
@@ -1,7 +1,0 @@
-# Wait until nginx is no longer restarting
-- name: Check for Restarting in docker ps
-  shell: docker ps | grep nginx
-  register: docker_nginx
-  until: docker_nginx.stdout.find("Restarting") == -1
-  delay: 1
-  retries: "{{ fifteen_minutes }}"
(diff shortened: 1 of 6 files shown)


playbooks/tasks/portals-check-nginx.yml
```diff
@@ -0,0 +1,7 @@
+# Wait until nginx is no longer restarting
+- name: Check for Restarting in docker ps
+  shell: docker ps | grep nginx
+  register: docker_nginx
+  until: docker_nginx.stdout.find("Restarting") == -1
```
kwypchlo, line 5: you can use inspect templating

```bash
docker inspect -f '{{print .State.Restarting}}' nginx
```

kwypchlo (changes requested):

kwypchlo:
> This PR adds the check that the NGINX docker container is up and not restarting. It has a 15min timeout.
> 
> I added a note saying that the check is temporary, needed for next week's deploy. But it should be an instant pass in normal cases so there is no real harm in leaving it in.

I think it would be better to leave it there - we will need it in case we add any more certificates and we need to wait until they generate. Also we will need this check in new server setup ansible job.

playbooks/tasks/portals-check-nginx.yml
firyx: Or you can use docker filters:
`ansible.builtin.command: docker ps --filter name=nginx --filter status=running`
and check you got 2 lines - header and data line.

firyx (changes requested):

playbooks/tasks/portals-check-nginx.yml
kwypchlo: > Or you can use docker filters:
> `ansible.builtin.command: docker ps --filter name=nginx --filter status=running`
> and check you got 2 lines - header and data line.
> 
> Anyway we should be aware of a possible issue while checking not `Restarting` or `running` status:
> Container might be in a `running` state like 100 ms before Nginx restarts and then move to `restarting` state in a loop.

is there a way to double check ? like have a condition where you need a success twice in a row with a delay of 1s for example

firyx (approved):

kwypchlo (approved):

PR #98, opened 2021-09-07
Generate env

firyx:
# PULL REQUEST

## Overview

Changes in this PR generate complete `.env` file even for cluster portals with accounts during execution `portal-setup-following` script.
Data are generated from `.env.j2` Jinja2 template and are loaded from ansible inventory, webportals.yml and from common, cluster and server configs stored in LastPass.

## Checklist
 - [ ] Changelog file created


playbooks/host_vars/crap.yml
```diff
@@ -3,7 +3,7 @@
 
 passworded_users:
   - username: "fil"
-    password_hash: "{{ lookup('community.general.lastpass', 'fil@crap.siasky.net', field='password') | password_hash('sha512', 65534 | random(seed=inventory_hostname) | string) }}"
+    password_hash: "{{ (lookup('community.general.lastpass', 'fil@crap.siasky.net', field='password') | password_hash('sha512', 65534 | random(seed=inventory_hostname) | string)) if (lastpass_required|default(False)) else None }}"
```
MSevey, line 6: nit: is it possible in ansible to define some vars to make this more readable? could the following be a variable to better understand it?
```
(lookup('community.general.lastpass', 'fil@crap.siasky.net', field='password') | password_hash('sha512', 65534 | random(seed=inventory_hostname) | string))
```

playbooks/tasks/lastpass-load-yaml-note.yml
```diff
@@ -0,0 +1,42 @@
+---
+
+# Load yaml secure note from LastPass to Ansible variable
```
MSevey, line 3: This is helpful 👍

playbooks/templates/.env.j2
```diff
@@ -1,118 +1,237 @@
-###############################################################################
-# Server specific
-
-# a portal domain name in a form <subdomain>.<domain>.<top-level-domain>
-DOMAIN_NAME={{ webportal_server_config.domain_name }}
-
-# a domain that needs to be set to a domain with tld ie. `siasky.net`
-# this is the portal domain that will be used for generating certificates and nginx routing
-PORTAL_DOMAIN={{ webportal_cluster_config.portal_cluster_domain if (portal_cluster_id is defined) else webportal_server_config.domain_name }}
+# xxxqqq: # xxx: Remove 'skynetlabs_devs_email' from:
```
MSevey, line 1: Are these notes meant to be left in?

playbooks/templates/.env.j2
```diff
@@ -1,118 +1,237 @@
...
-PORTAL_NAME='{{ webportal_server_config.portal_name }}'
+###############################################################################
+# Server specific
 
-# a list of active portal modules. e.g. 'PORTAL_MODULES=aj'. Possible choices:
-# 'a': accounts, 'j': Jaeger
+# A portal domain name
+# For a single server: single-portal.com
+# For a cluster: server-1.cluster.com
+# xxx: Is this stil used somewhere?
```
MSevey, line 26: Did you get an answer on this?

playbooks/templates/.env.j2
```diff
@@ -1,118 +1,237 @@
...
+SIA_WALLET_PASSWORD={{ webportal_server_config.sia_wallet_password if (webportal_server_config.sia_wallet_password is defined and webportal_server_config.sia_wallet_password is not none and webportal_server_config.sia_wallet_password != '') else (webportal_server_config.sia_wallet_seed if webportal_server_config.sia_wallet_seed is defined else '') }}
 
-# certificate string that will be used to generate ssl certificates, read more in docker/caddy/Caddyfile
-SSL_CERTIFICATE_STRING={{ webportal_cluster_config.portal_cluster_domain + ', *.' + webportal_cluster_config.portal_cluster_domain + ', *.hns.' + webportal_cluster_config.portal_cluster_domain if (webportal_cluster_config is defined) else webportal_server_config.domain_name }}, *.{{ webportal_server_config.domain_name }}, *.hns.{{ webportal_server_config.domain_name }}
+# Certificate string that will be used to generate ssl certificates, read more in docker/caddy/Caddyfile
+# xxxqqq: Deprecated
+{% if portal_cluster_id is defined %}
+SSL_CERTIFICATE_STRING={{ cluster_domain }}, *.{{ cluster_domain }}, *.hns.{{ cluster_domain }}, *.{{ server_domain }}, *.hns.{{ server_domain }}
```
MSevey, line 73: Should we just drop this now?  The update to not need it is being deployed. I

playbooks/templates/.env.j2
```diff
@@ -1,118 +1,237 @@
...
+# For a cluster: https://cluster.com
+SKYNET_PORTAL_API=https://{{
+    cluster_domain
+    if portal_cluster_id is defined
+    else server_domain
+}}
+
+# Administrator contact email.
+# Needed for communication regarding SSL certification.
+# xxxqqq: Deprecated.
```
MSevey, line 98: So can we just remove it then?

playbooks/templates/.env.j2
```diff
@@ -1,118 +1,237 @@
...
 EMAIL_ADDRESS={{ skynetlabs_devs_email }}
 
-# (optional) if using cloudflare as dns loadbalancer (need to change it in
-# Caddyfile too)
+# (optional) If using CloudFlare as DNS loadbalancer (need to change it in
+# Caddyfile too).
+# xxx: Not implemented in Ansible
```
MSevey, line 103: nit: `TODO` is more common than `xxx` to know that this is something that needs a follow up.

playbooks/templates/.env.j2
```diff
@@ -1,118 +1,237 @@
...
 CLOUDFLARE_AUTH_TOKEN=
 
-# (optional) if using route53 as a dns loadbalancer
+# (optional) If using Route53 as a DNS loadbalancer.
```
MSevey, line 106: Can we group these into a `DNS` section? That might be a little clearer for community operators.

playbooks/templates/.env.j2
```diff
@@ -1,118 +1,237 @@
...
+# (optional) Only required if you're using the Discord notifications
+# integration.
+# xxx move those to common/cluster configs in LastPass
 DISCORD_WEBHOOK_URL={{ discord_webhook_url | default('') }}
 DISCORD_MENTION_ROLE_ID={{ discord_mention_role_id | default('') }}
 DISCORD_MENTION_USER_ID={{ discord_mention_user_id | default('') }}
 
 # common path for cluster, single path for each dev server
-S3_BACKUP_PATH={{ webportal_common_config.s3_backup_path if ('webportals_prod' in group_names) else (webportal_common_config.s3_backup_path + '_' + inventory_hostname) }}
+# S3 backup path.
```
MSevey, line 120: @ro-tex the S3 backup is for mongo right? Anything else?

We should add to the comment what S3 is being used for.

MSevey: Then I would also list it as optional and we should link to what is triggering the backups. 

Fine for a follow up, just thinking through questions a community portal operator might have. They might want to back it up somewhere else and they should be able to.

playbooks/templates/.env.j2
```diff
@@ -1,118 +1,237 @@
...
 
-# xxx
-# AIRTABLE_API_KEY=
-# SQUARE_ACCESS_TOKEN=
-# STRIPE_PUBLISHABLE_KEY=
-# STRIPE_API_KEY=
-# STRIPE_SECRET_KEY=
-# STRIPE_WEBHOOK_SECRET=
+# Used to load skylinks blocklist data from Airtable
+AIRTABLE_API_KEY={{ webportal_common_config.airtable_api_key }}
```
MSevey, line 126: Same here, this is optional and community portal operations should be able to use something else. That will require some changes to the code in terms of allowing for other locations.

Can be a follow up.

playbooks/templates/.env.j2
```diff
@@ -1,118 +1,237 @@
...
+
+# MongoDB address or container name
 SKYNET_DB_HOST=mongo
 
-# this is the MongoDB port
+# MongoDB port
 SKYNET_DB_PORT=27017
 
-# this is accounts log level
+# Accounts log level
```
MSevey, line 183: I would add a note saying that this is for Jaeger.

MSevey (approved):
I left a bunch a minor thoughts and questions. Nothing blocking. All fine to be handled in follow ups.

playbooks/tasks/lastpass-load-webportal-config.yml
```diff
@@ -1,12 +1,32 @@
...
-# Load Portal server and common configs from LastPass
+# Load portal configs from LastPass
 
-- name: Include loading LastPass portal server config
-  include_tasks: tasks/lastpass-load-webportal-server-config.yml
+# Portal common and server configs are loaded and set always.
+# If a portal is part of a cluster:
+# - cluster config is loaded and set
+# - cluster config data are merged into and can override common config data
+# - common config then contans valid common and cluster data
```
ro-tex, line 9: This is a little confusing. If I'm reading this correctly, then the configuration is split into two parts - `common` config (shared by all servers) and `server` config (per server).

My question is: Shouldn't this setup be used only when the portal is using a cluster of servers? Otherwise the `common` and `server` configs will be identical, right?

In that case isn't the `cluster` config the same as the `common` config? Or do we have all three?

Example:
`common` says "A: 1"
`cluster` says "A: 2" -> overrides `common` and sets A to 2
`server` says "A: 3" -> overrides all and sets A to 3

playbooks/templates/.env.j2
ro-tex: It's for Mongo and Cockroach right now but it will be just Mongo once we get rid of Kratos.

I agree that it should be marked as optional and that we should link to an explanation about backups.

playbooks/templates/.env.j2
ro-tex: It is not. This sets the log level for accounts and has nothing to do with Jaeger.

ro-tex:
What happened to `PORTAL_URL`? I use that in `accounts` in order to determine the issuer of the JWT. I also default to `https://siasky.net` but there should be a way for the user to override that. Should I use `PORTAL_DOMAIN` instead?

ro-tex (approved):

playbooks/host_vars/crap.yml
firyx: Created a f/u issue #99

playbooks/templates/.env.j2
firyx: Created f/u issue #100

playbooks/templates/.env.j2
firyx: As discussed online, will test portals without it and remove it if unused.
Issue #101

playbooks/templates/.env.j2
firyx: Will remove it after 2nd successful deploy. Issue #102

playbooks/templates/.env.j2
firyx: Will remove it after 2nd successful deploy. Issue #102

playbooks/tasks/lastpass-load-webportal-config.yml
firyx: There will be
* a generic portal setup Ansible role (WIP)
* a Skynet Labs specific portal setup Ansible playbook using this role

Generic role requires (implementation is upto operator, they do not have to use LastPass)
* generic load `common` config var and `server` config var
* generic save `server` config var

Skynet Labs specific playbook loads and saves to LastPass.

SkynetLabs has 3 types configs in LastPass:
* common - 1x
* cluster - currently 2x: prod, dev-a
* server - for each server
Each LastPass config has different data, they do not overlap.

In Skynet Labs Ansible playbook we have 3 config vars:
* common
* cluster
* server

Skynet Labs Ansible playbook does for each host (portal):
* `common` var loads `common` data from LastPass
* `cluster` var loads `cluster` data from LastPass
* `common` var is set to have `common` data overriden with `cluster` data from `cluster` var
* `cluster` var is set to null, because we have now merged data in `common` var for a specific server and Ansible role doesn't use `cluster` var
* `server` var loads `server` data from LastPass (and doesn't override anything)

We can change common settings by just editing `common` or `cluster` configs in LastPass and (WIP) rerunning portal setup script and we are sure they will be set on all servers, no server can override them by its own `server` data from LastPass.

Operator with just 1 server needs 1x `common` and 1x `server` configs data in their backend, they will not overlap.

ro-tex: OK. I was missing the part where the data don't overlap.

playbooks/templates/.env.j2
firyx: Created issue #103

playbooks/templates/.env.j2
firyx: Created issue #104

PR #107, opened 2021-09-13
Add variable to set number of parallel deployments

firyx:
# PULL REQUEST

## Overview

Till now we have performed deployments sequentially (rolling updates) one server at a time.

This PR adds an optional variable `parallel_deploys` that can set number of servers to deploy to in parallel.

## Checklist
 - [ ] Changelog file created

### DIFF
--- a/playbooks/portals-deploy.yml
+++ b/playbooks/portals-deploy.yml
@@ -11,7 +11,7 @@
   any_errors_fatal: True
 
   vars:
-    max_hosts: "{{ groups['webportals'] | length - 1 }}"
+    max_hosts: "{{ groups['webportals_prod'] | length - 1 }}"
 
     # Default batch config (one batch with all selected hosts will be created),
     # can be overriden with cli flag --extra-vars/-e
--- a/playbooks/portals-deploy.yml
+++ b/playbooks/portals-deploy.yml
@@ -2,19 +2,27 @@
   hosts: webportals
   gather_facts: False
 
-  # Limit concurrency to keep batch_hosts order deterministic
+  # Limit concurrency to keep batch_hosts order deterministic.
+  # When you want to deploy in parallel, keep 'serial: 1' for this play,
+  # parallel deploys are set by 'serial' variable in the next play.
   serial: 1
 
   # Stop on first error, do not execute on the next host
   any_errors_fatal: True
 
   vars:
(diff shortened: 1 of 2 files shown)


playbooks/portals-deploy.yml
```diff
@@ -28,30 +36,25 @@
   gather_facts: False
 
   # Limit concurrency
-  serial: 1
+  serial: "{{ parallel_deploys | default(1) }}"
```
ro-tex, line 39: Can we rename this var to `parallel`? `serial` sounds like a boolean flag while `parallel` sounds like something that controls how many parallel streams of deployments we have. Which functions the same way for `parallel: 1`.

MSevey: I don't think so. Fil can confirm but `serial` is an ansible builtin, not something we are defining.

ro-tex: I also don't quite understand the interaction between the two `serial` vars. When would we want to set the first `serial` to something other than 1?

MSevey (approved):

playbooks/portals-deploy.yml
```diff
@@ -2,19 +2,27 @@
...
+    max_hosts: "{{ groups['webportals'] | length - 1 }}"
+
     # Default batch config (one batch with all selected hosts will be created),
     # can be overriden with cli flag --extra-vars/-e
     batch_size: 1
     batch_number: 1
 
   tasks:
+    # Check '--limit' is used
+    - name: Fail if you are targeting all dev and prod webportals
```
ro-tex, line 23: It only now occurs to me that if we're targeting all prod portals this won't fail. Let's change it to fail when we target all prod portals (or at least as many portals as we have in the prod group).

playbooks/portals-deploy.yml
ro-tex: > I don't think so. Fil can confirm but `serial` is an ansible builtin, not something we are defining.

Oh, that would make sense.

firyx: `serial` is a built-in Ansible variable. Our name is `parallel_deploys`.

playbooks/portals-deploy.yml
firyx: Updated.

ro-tex (approved):

MSevey (approved):

PR #114, opened 2021-09-14
refactor out of template playbook

MSevey:
# PULL REQUEST

## Overview
To help with others getting started with Ansible, I added a little template playbook file. 

## Checklist
 - [x] Changelog file created

### DIFF
--- a/playbooks/templates/playbook.yml
+++ b/playbooks/templates/playbook.yml
@@ -1,0 +1,51 @@
+# Name your playbook, this is printed in STDOUT
+- name: My Cool Playbook
+  # Define your target hosts, this typically doesn't change
+  hosts: webportals
+  # Define the user that is executing the commands, this typically doesn't
+  # change
+  remote_user: "{{ webportal_user }}"
+  # Tell ansible if it should gather information about the hosts. This is
+  # currently only used in setting up the servers.
+  gather_facts: False
+
+  # Limit concurrency or the playbook if needed
+  serial: 1
+
+  # Execute playbook on hosts in parallel as fast as possible (do not wait for other hosts)
+  # strategy: free
+
+  # Stop on first error, do not execute on the next host
+  any_errors_fatal: True
+
+  # Playbook specific vars
+  vars:
+    max_hosts: "{{ groups['webportals'] | length - 1 }}"
+
+  # Define the tasks for your playbook
+  tasks:
+    - name: Hello World
+      ansible.builtin.command: echo "Hello World"
+
+    # Example of calling a refactored task
+    #
+    # Check '--limit' is used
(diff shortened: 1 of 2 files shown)


playbooks/templates/playbook.yml
```diff
@@ -0,0 +1,51 @@
...
+    #
+    # Check '--limit' is used
+    - name: Fail if you are targeting all dev and prod webportals
+      include_tasks: tasks/host-limit-check.yml
+
+    # Example of task with additional vars
+    #
+    # Update log status to 'tested'
+    - name: Update log status to 'tested'
+      include_tasks: tasks/portal-logs-update-status.yml
```
firyx, line 40: If this playbook/task is run against a portal host, it might set portal status file incorrectly, this can spoil possible rollback.

playbooks/templates/playbook.yml
```diff
@@ -0,0 +1,51 @@
...
+        tag_from: "started"
+        tag_to: "tested"
+
+    # Example of task with 'when' condition
+    #
+    # Enable health check
+    - name: Include enabling portal health check
+      # do not enable health checks on hosts from out_of_LB group
+      when: ('out_of_LB' not in group_names)
+      include_tasks: tasks/portal-health-check-enable.yml
```
firyx, line 51: Again: Running this can make undesired change to the host.
I would consider including tasks, that do not change state of the host.

firyx (changes requested):

playbooks/templates/playbook.yml
MSevey: The idea is that this is just an example.  I'm not suggesting that all playbooks would use these tasks.

playbooks/templates/playbook.yml
MSevey: same answer.

playbooks/templates/playbook.yml
firyx: I understand the idea, but we should use tasks that do not interfere with our fleet when somebody tries to run the template playbook(s).

firyx (approved):

ro-tex (approved):

playbooks/templates/playbook.yml
```diff
@@ -0,0 +1,45 @@
...
+    max_hosts: "{{ groups['webportals'] | length - 1 }}"
+
+  # Define the tasks for your playbook
+  tasks:
+    - name: Hello World
+      ansible.builtin.command: echo "Hello World"
+
+    # Example of calling a refactored task
+    - name: Run my refactored task
+      include_tasks: tasks/my-factored-task.yml
```
ro-tex, line 32: Should we make sure this task exists?

PR #117, opened 2021-09-17
Add playbook and task for docker command

MSevey:
# PULL REQUEST

## Overview
There are often times that we want to execute a docker command on all the portals so I created a little helper playbook to do so. I created a task so that it could also be included in deployment when needed.

## Checklist
 - [x] Changelog file created


playbooks/portals-docker-command.yml
```diff
@@ -0,0 +1,55 @@
+- name: Run Docker Command on Skynet Webportal
+  hosts: webportals
+  remote_user: "{{ webportal_user }}"
+  gather_facts: False
+
+  # Target 3 servers at a time.
+  serial: 3
```
MSevey, line 7: Follow Up: we should refactor `parallel_deploys` to be a common `num_parallel_hosts` for all playbooks to use.

firyx (approved):

playbooks/portals-docker-command.yml
```diff
@@ -0,0 +1,55 @@
...
+
+  # Playbook specific vars
+  vars:
+    max_hosts: "{{ groups['webportals'] | length - 1 }}"
+
+  tasks:
+    # Check '--limit' is used. This is to protect against taking the entire
+    # fleet down
+    - name: Fail if you are targeting all dev and prod webportals
+      include_tasks: tasks/host-limit-check.yml
```
ro-tex, line 20: Do we want to have this check for this one?

playbooks/portals-docker-command.yml
```diff
@@ -0,0 +1,55 @@
...
+
+      when: "'webportals_takedown' in group_names"
+
+    # Check/install Ansible prerequisites
+    - name: Include preparing portal prerequisites
+      include_tasks: tasks/portals-prepare.yml
+
+    # Disable health check
+    - name: Include disabling portal health check
+      include_tasks: tasks/portal-health-check-disable.yml
```
ro-tex, line 39: This could be optional. The reason for that is that sometimes we will want to do something trivial like change a config value on all servers that won't take effect until the server is restarted. We don't need to add extra stress to the network by disabling servers.

ro-tex (approved):

playbooks/portals-docker-command.yml
MSevey: I disagree. That adds a human decision into the process which isn't scalable. 
We should handle all events the same way regardless of size. Something might seem trivial at the time but have an unintended consequence.

playbooks/portals-docker-command.yml
MSevey: @ro-tex  yes, this check should always be in. This prevents someone from accidentally running a command against all possible servers, production, dev, staging, etc.

PR #137, opened 2021-10-13
490 Handling for disabling health check and MongoDB compose update

MSevey:
# PULL REQUEST

## Overview
Add handling for case when siac is not running for disabling the health checks. Also adds support for the `m` option for `PORTAL_MODULES`.

Tested on eu-fin-1

## Checklist
 - [x] Changelog file created

### DIFF
--- a/playbooks/group_vars/webportals.yml
+++ b/playbooks/group_vars/webportals.yml
@@ -115,10 +115,14 @@ webportal_setup_health_checks: True
 webportal_docker_compose_files_mandatory:
   - "docker-compose.yml"
 
+# Docker compose files should be in order of dependency. For example, accounts
+# depends on mongo so mongo needs to be before accounts in case accounts
+# overrides any mongo settings. This is also why the override file should always
+# be last.
 webportal_docker_compose_files_optional:
+  - "docker-compose.mongodb.yml"
   - "docker-compose.accounts.yml"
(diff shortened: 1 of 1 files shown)


playbooks/group_vars/webportals.yml
```diff
@@ -125,6 +125,7 @@ webportal_docker_compose_files_optional:
 webportal_docker_compose_files_dict:
   a: "docker-compose.accounts.yml"
   j: "docker-compose.jaeger.yml"
+  m: "docker-compose.mongodb.yml"
```
firyx, line 128: We need to add `docker-compose.mongodb.yml` also to `webportal_docker_compose_files_optional` list above.
(Didn't find a way to do it via suggestion.)

firyx (changes requested):

playbooks/group_vars/webportals.yml
```diff
@@ -118,13 +118,15 @@ webportal_docker_compose_files_mandatory:
 webportal_docker_compose_files_optional:
   - "docker-compose.accounts.yml"
   - "docker-compose.jaeger.yml"
+  - "docker-compose.mongodb.yml"
```
firyx, line 121: mongo should be moved before accounts, because accounts depend on mongo and can possibly override it.

PR #155, opened 2021-11-04
Add check for keeping MongoDB replicaset majority online

firyx:
# PULL REQUEST

## Overview

Currently there is a risk that we stop portal mongo services during deploys, takedowns, etc. and there will be only 50% or less of MongoDB voting members online, which will disrupt portal operation.

This PR adds a check before stoping mongo service to be sure we keep > 50% MongoDB voting members online during execution of our playbooks.

## Checklist
 - [x] Changelog file created

### DIFF
--- a/playbooks/tasks/portal-mongo-replicaset-check-majority.yml
+++ b/playbooks/tasks/portal-mongo-replicaset-check-majority.yml
@@ -35,18 +35,20 @@
               .filter(m => m.priority > 0)
               .map(m => m.host)
             onlineVotingMembers = rs.status().members
-              .filter(m => m.stateStr == 'PRIMARY' || m.stateStr == 'SECONDARY')
+              .filter(m => (m.stateStr == 'PRIMARY' || m.stateStr == 'SECONDARY') && m.health == 1)
               .map(m => m.name)
               .filter(m => votingMembers.includes(m))
             batchMembers = [{% for h in ansible_play_batch -%}
               '{{ hostvars[h].ansible_host + ':' + (hostvars[h].subcrap_mongo_port | default(default_mongo_port) | string) }}'{{ loop.last | ternary('', ', ') }}
               {%- endfor -%}]
             onlineVotingMembersAfter = onlineVotingMembers.filter(m => !batchMembers.includes(m))
+            thisMemberIsVoting = votingMembers.includes('{{ ansible_host + ':' + (subcrap_mongo_port | default(default_mongo_port) | string) }}')
             result = {
               "votingMembers": votingMembers,
               "onlineVotingMembers": onlineVotingMembers,
               "batchMembers": batchMembers,
               "onlineVotingMembersAfter": onlineVotingMembersAfter,
+              "thisMemberIsVoting": thisMemberIsVoting,
               "replicasetMajorityOk": onlineVotingMembersAfter.length > votingMembers.length/2
             }
           until: True
@@ -64,7 +66,7 @@
           - Voting members in config: {{ mongo_replicaset_result.votingMembers | length }}
             {{ mongo_replicaset_result.votingMembers }}
 
-          - Voting members currently online: {{ mongo_replicaset_result.onlineVotingMembers | length }}
+          - Voting members currently online and healthy: {{ mongo_replicaset_result.onlineVotingMembers | length }}
             {{ mongo_replicaset_result.onlineVotingMembers }}
 
           - Members in the current Ansible batch: {{ mongo_replicaset_result.batchMembers | length }}
@@ -81,6 +83,7 @@
           - Don't stop so many voting members in one Ansible batch
           - Reconfigure voting members (move voting from an offline member to some online member)
       when: >-
+        mongo_replicaset_result.thisMemberIsVoting and
         mongo_replicaset_result.votingMembers | length > 2 and
         not mongo_replicaset_result.replicasetMajorityOk
   when:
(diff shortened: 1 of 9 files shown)


playbooks/tasks/portal-mongo-replicaset-check-majority.yml
```diff
@@ -0,0 +1,89 @@
...
+          skynet_db_user: "{{ skynet_db_user_result.stdout }}"
+          skynet_db_pass: "{{ skynet_db_pass_result.stdout }}"
+
+    - name: Include checking mongo replicaset majority
+      include_tasks: tasks/portal-role-task-mongo-shell-eval.yml
+      vars:
+        mongodb_shell:
+          eval: |
+            votingMembers = rs.config().members
+              .filter(m => m.priority > 0)
```
ro-tex, line 35: REQ:
```suggestion
              .filter(m => m.votes > 0)
```

playbooks/tasks/portal-mongo-replicaset-check-majority.yml
```diff
@@ -0,0 +1,89 @@
...
+              .map(m => m.host)
+            onlineVotingMembers = rs.status().members
+              .filter(m => m.stateStr == 'PRIMARY' || m.stateStr == 'SECONDARY')
+              .map(m => m.name)
+              .filter(m => votingMembers.includes(m))
```
ro-tex, line 40: This should also filter by `health=1`, otherwise we'll be counting crashed nodes as online.

playbooks/tasks/portal-mongo-replicaset-check-majority.yml
```diff
@@ -0,0 +1,89 @@
...
+              "replicasetMajorityOk": onlineVotingMembersAfter.length > votingMembers.length/2
+            }
+          until: True
+          retries: 0
+
+    - name: Set mongo replicaset result
+      set_fact:
+        mongo_replicaset_result: "{{ mongo_shell_result.transformed_output | from_json }}"
+
+    - name: Stop the playbook if we don't have majority (> 50%) MongoDB replicaset voting members online
```
ro-tex, line 59: The first thing this playbook should do is check whether the current node is a voting member or not. If it's not a voting member then we don't need to check anything, we can just proceed.

ro-tex (changes requested):

playbooks/tasks/portal-mongo-replicaset-check-majority.yml
firyx: Added.

playbooks/tasks/portal-mongo-replicaset-check-majority.yml
firyx: Updated fail condition not to fail for non-voting member.

ro-tex (approved):

playbooks/tasks/portal-role-task-mongo-check-docker-service.yml
```diff
@@ -83,6 +83,7 @@
       eval: "db.serverStatus()"
       until: "'\"ok\" : 1' in mongo_shell_result.transformed_output"
       retries: 20
+      ignore_errors: True
   when: mongo_service_ok
   register: mongo_checking_admin_result
```
MSevey, line 88: Since we are ignoring errors here, shouldn't we be adding an `is defined` check for where this variable is being used to avoid undefined errors?

playbooks/tasks/portal-role-task-mongo-check-docker-service.yml
```diff
@@ -113,6 +114,7 @@
         })
       until: mongo_shell_result.transformed_output in ['PRIMARY', 'SECONDARY']
       retries: 120
+      ignore_errors: True
   when: mongo_service_ok
   register: checking_replicaset_result
```
MSevey, line 119: same

playbooks/tasks/portal-role-task-mongo-check-docker-service.yml
MSevey: I link line 102

playbooks/tasks/portal-role-task-mongo-check-docker-service.yml
MSevey: I think line 133

MSevey (changes requested):
I think there are two potential undefined edge cases.

playbooks/tasks/portal-role-task-mongo-check-docker-service.yml
firyx: `ignore_errors: True` in this case and below
do not belong to `inlcude_task` task like this:

```yaml
include_tasks: tasks/portal-role-task-mongo-shell-eval.yml
ignore_errors: True
register: mongo_checking_admin_result
```

They are passed down (to a task in `tasks/portal-role-task-mongo-shell-eval.yml`) via `mongodb_shell.ignore_errors` to

```yaml
community.mongodb.mongodb_shell:
register: mongo_shell_result
ignore_errors: "{{ mongodb_shell.ignore_errors | default(False) }}"
```

The result which is affected is not `mongo_checking_admin_result` or `checking_replicaset_result`
but `mongo_shell_result` and it is handled with:

`"{{ not (mongo_shell_result.failed | default(False)) }}"` in both cases.

MSevey (approved):

PR #161, opened 2021-11-12
Block ip ranges

firyx:
# PULL REQUEST

## Overview

Block malicious IP ranges with ufw firewall. The list of IP ranges is stored in `ansible-private` repository.

## Checklist
 - [x] Changelog file created

### DIFF
--- a/README.md
+++ b/README.md
@@ -501,8 +501,10 @@ private variables are stored. Default path is in `ansible-private` repository,
 in `private-vars/private-vars.yml` file. In this file there are defined 2 lists
 of IPs or Ip ranges: `incoming_ips_ip_ranges_block` and
 `incoming_ips_ip_ranges_unblock`. To block an IP/IP range, add it to the block
-list, to unblock the previously blocked IP/IP range, remove it from the block
-list and add it to the unblock list and execute the playbook.
+list, execute the playbook and keep the IP/IP range in the list so it is
+blocked also later on newly setup portals. To unblock the previously blocked
+IP/IP range, remove it from the block list and add it to the unblock list and
+execute the playbook.
 
 Example `private-vars.yml` file:
 ```yaml
--- a/README.md
+++ b/README.md
@@ -30,6 +30,7 @@
         - [Set Allowance Max Storage Price](#set-allowance-max-storage-price)
         - [Block Portal Skylinks](#block-portal-skylinks)
         - [Unblock Portal Skylinks](#unblock-portal-skylinks)
+        - [Block and Unblock Incoming Traffic to Portals](#block-and-unblock-incoming-traffic-to-portals)
         - [Run Integration Tests](#run-integration-tests)
         - [Run Health Checks](#run-health-checks)
         - [Setup Portal from Scratch](#setup-portal-from-scratch)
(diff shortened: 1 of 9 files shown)


ro-tex (dismissed):

playbooks/portals-setup-following.yml
```diff
@@ -27,7 +27,7 @@
           become: True
     
     - name: Include firewall setup
-      include_tasks: tasks/portal-firewall-ufw-setup.yml
```
ro-tex, line 30: We are replacing this one, should we also replace the one in `portal-setup-ufw.yml` playbook, or that one is fine?

README.md
```diff
@@ -485,6 +486,50 @@ To run:
...
+  `private_vars.incoming_ips_ip_ranges_unblock` list.
+
+Preparation:  
+`private_vars_file` defined in `webportals.yml` defines a filepath where
+private variables are stored. Default path is in `ansible-private` repository,
+in `private-vars/private-vars.yml` file. In this file there are defined 2 lists
+of IPs or Ip ranges: `incoming_ips_ip_ranges_block` and
+`incoming_ips_ip_ranges_unblock`. To block an IP/IP range, add it to the block
+list, to unblock the previously blocked IP/IP range, remove it from the block
+list and add it to the unblock list and execute the playbook.
```
ro-tex, line 505: So, if I want to block a new range I need to add it to the list. After the range is blocked, we would generally remove it from the list, then?

I wonder if it wouldn't be better to keep all the blocked ranges in that list, so we can easily apply them to newly created servers. This could make the list very very long and hard to work with, though.

Can we relatively easily sync up all machines on the list of IP ranges they block?

ro-tex (approved):
LGTM.

Two questions/suggestions but they are not blocking.

playbooks/portals-setup-following.yml
firyx: In `playbooks/portals-setup-following.yml`
we call `tasks/portal-firewall-setup.yml`
which calls both:
- `tasks/portal-firewall-iptables-block-unblock.yml`
- `tasks/portal-firewall-ufw-setup.yml`

We have 2 now firewall playbooks:
- `playbooks/portals-block-unblock-incoming-traffic.yml` - for un-/blocking incoming traffic in iptables
- `playbooks/portal-setup-ufw.yml` - for the rest of firewall done via ufw
so we can target both individually.

README.md
firyx: > So, if I want to block a new range I need to add it to the list. After the range is blocked, we would generally remove it from the list, then?
> 

No, we add the range to the list, we block it and we keep it in the list.

We remove the range from the block list and add it to unblock list only in case we want to unblock it later.

> I wonder if it wouldn't be better to keep all the blocked ranges in that list, so we can easily apply them to newly created servers. This could make the list very very long and hard to work with, though.
> 

Yes, we keep blocked ranges in the list, so they are applied also to newly created servers.

> Can we relatively easily sync up all machines on the list of IP ranges they block?

See above.

ro-tex (approved):

MSevey (approved):

PR #166, opened 2021-11-24
Add playbook to update all apt packages

firyx:
# PULL REQUEST

## Overview

Update all apt packages (take the portal out of loadbalancer before, run integration tests, health check after).

## Checklist
 - [x] Changelog file created

### DIFF
--- a/playbooks/x-portals-ensure-latest-packages.yml
+++ b/playbooks/x-portals-ensure-latest-packages.yml
@@ -1,0 +1,67 @@
+- name: Ensure Latest Docker is Installed
+  hosts: webportals
+  remote_user: "{{ webportal_user }}"
+  gather_facts: True
+
+  # Limit concurrency
+  serial: 1
+
+  # Stop on first error, do not execute on the next host
+  any_errors_fatal: True
+
+  # Playbook specific vars
+  vars:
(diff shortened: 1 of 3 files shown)


ro-tex (approved):

playbooks/x-portals-ensure-latest-packages.yml
```diff
@@ -0,0 +1,66 @@
...
+    - name: Include disabling health check and stopping portal docker services
+      include_tasks: tasks/portal-stop.yml
+
+    - name: Perform apt-get dist-upgrade to update all apt packages
+      ansible.builtin.apt:
+        upgrade: dist
+        update_cache: True
+      become: True
+
+    - block:
```
MSevey, line 33: Does having this block here leave the portal completely down? skyd would have been stopped in `tasks/portal-stop.yml` and nothing has brought it back up yet.

Probably should bring it all up, do this tests, and then take it down. This would ensure that the portal continues to be fully functioning and we don't miss updates when we eventually bring it back up.

ro-tex (approved):
LGTM aside from the active discussion.

firyx:
If we need this functionality, we should implement it inside `portals-restart` with an optional flag to be set in `config.yml` to enable this.

PR #168, opened 2021-11-24
Add playbook for gzipping logs

MSevey:
# PULL REQUEST

## Overview
Since logrotate never worked and we keep seeing disk space issues I made a helper ansible script that we can use to gzip various logs. The log files are defined with their absolute paths in the `portal_log_files` variable, so we can easily add logs. The current size trigger is 5GB which can easily be updated as well.

## Example for Visual Changes

Initial test on dev1 with 1GB as the limit
Screen Shot 2021-11-24 at 1 14 42 PM

Verified on dev3
Screen Shot 2021-11-24 at 1 42 45 PM

Screen Shot 2021-11-24 at 1 44 06 PM

## Checklist
 - [x] Changelog file created

### DIFF
--- a/playbooks/group_vars/webportals.yml
+++ b/playbooks/group_vars/webportals.yml
@@ -180,11 +180,11 @@ private_vars_file: "../../../ansible-private/private-vars/private-vars.yml"
 
 # Portal log file paths
 portal_log_files:
-  - "/home/user/skynet-webportal/docker/data/sia/renter/contractor.log"
-  - "/home/user/skynet-webportal/docker/data/sia/renter/hostdb.log"
-  - "/home/user/skynet-webportal/docker/data/sia/renter/renter.log"
-  - "/home/user/skynet-webportal/docker/data/sia/renter/repair.log"
-  - "/home/user/skynet-webportal/docker/data/sia/siamux/siamux.log"
+  - "{{ webportal_dir }}/docker/data/sia/renter/contractor.log"
+  - "{{ webportal_dir }}/docker/data/sia/renter/hostdb.log"
+  - "{{ webportal_dir }}/docker/data/sia/renter/renter.log"
+  - "{{ webportal_dir }}/docker/data/sia/renter/repair.log"
+  - "{{ webportal_dir }}/docker/data/sia/siamux/siamux.log"
 
 # The size limit at which log files will be gzipped 1GB
 # 1GB 1073741824
--- a/playbooks/group_vars/webportals.yml
+++ b/playbooks/group_vars/webportals.yml
@@ -177,3 +177,16 @@ docker_services_exclude_waiting:
 local_logs_dir: "../my-logs"
 # Path relative to ./playbooks/vars directory
 private_vars_file: "../../../ansible-private/private-vars/private-vars.yml"
+
+# Portal log file paths
+portal_log_files:
+  - "/home/user/skynet-webportal/docker/data/sia/renter/contractor.log"
+  - "/home/user/skynet-webportal/docker/data/sia/renter/hostdb.log"
+  - "/home/user/skynet-webportal/docker/data/sia/renter/renter.log"
+  - "/home/user/skynet-webportal/docker/data/sia/renter/repair.log"
+  - "/home/user/skynet-webportal/docker/data/sia/siamux/siamux.log"
+
+# The size limit at which log files will be gzipped 1GB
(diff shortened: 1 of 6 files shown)


DavidVorick:
I don't think we need to keep most of these logs. Can you keep the nginx` access.log` files out of the script? I'll set up a separate thing to compress them as part of my metrics manager.

playbooks/tasks/portal-gzip-logs.yml
```diff
@@ -0,0 +1,24 @@
...
+
+# Stat all the portal_log_files.
+- name: Stat Files
+  ansible.builtin.stat:
+    path: "{{ file }}"
+  register: loop_results
+  loop: "{{ portal_log_files }}"
+  loop_control:
+    loop_var: file
+  when: portal_log_files is defined
```
firyx, line 11: I would leave out this `when`. If `portal_log_files` are not defined user gets a message of the missing config which should be mandatory for this playbook.

playbooks/tasks/portal-gzip-logs.yml
```diff
@@ -0,0 +1,24 @@
...
+  become: True
+
+# GZIP any files over the portal_log_size_gzip_limit
+- name: GZIP Files
+  community.general.archive:
+    path: "{{ file.stat.path }}"
+    dest: "{{ file.stat.path + lookup('pipe','date +%Y-%m-%dT%H:%M:%S') + '.gz' }}"
+    remove: yes
+  loop: "{{ loop_results.results | from_yaml | list }}"
```
firyx, line 20: nit: I think `| from_yaml | list` is not needed, `loop_results.results` should be already a list.

firyx (changes requested):

playbooks/portals-gzip-logs.yml
```diff
@@ -0,0 +1,77 @@
...
+      include_tasks: tasks/portal-health-checks-run.yml
+
+    # Takedown portals in takedown group
+    - block:
+        # Include getting wanted docker-compose files
+        - name: Include getting wanted docker-compose files
+          include_tasks: tasks/portal-docker-compose-files-get-wanted.yml
+
+        # Start just sia docker service
+        - name: Start just sia docker service
```
firyx, line 56: Sia is already running, should this be moved up?
Also skyd now requires mongo.

firyx (changes requested):

playbooks/tasks/portal-gzip-logs.yml
```diff
@@ -0,0 +1,24 @@
...
+  register: loop_results
+  loop: "{{ portal_log_files }}"
+  loop_control:
+    loop_var: file
+  when: portal_log_files is defined
+  become: True
+
+# GZIP any files over the portal_log_size_gzip_limit
+- name: GZIP Files
+  community.general.archive:
```
firyx, line 16: Seems to need `become: True` to modify existing log files.

firyx (changes requested):

playbooks/tasks/portal-gzip-logs.yml
MSevey: it is a list, but not a nicely formatted one. This was from a StackOverflow post.  If you try and loop over it and print it out it is all funky,

playbooks/group_vars/webportals.yml
```diff
@@ -177,3 +177,16 @@ docker_services_exclude_waiting:
 local_logs_dir: "../my-logs"
 # Path relative to ./playbooks/vars directory
 private_vars_file: "../../../ansible-private/private-vars/private-vars.yml"
+
+# Portal log file paths
+portal_log_files:
```
firyx, line 182: Should use:
`- "{{ webportal_dir }}/docker/data/..."`

firyx (changes requested):

firyx (changes requested):
We should make gzipping logs optional part of portal deploy or portal restart, same as `clean_renter_log` is used here:
https://github.com/SkynetLabs/ansible-playbooks/blob/master/playbooks/tasks/portal-stop.yml#L14

Gzipping should be done via Alpine docker container so we don't need to add sudo (LastPass login) to portal restart/deploy playboks like here:
https://github.com/SkynetLabs/ansible-playbooks/blob/master/playbooks/tasks/portal-logs-clean-renter-log.yml

PR #178, opened 2021-12-15
Address some error cases

MSevey:
# PULL REQUEST

## Checklist
 - [ ] Changelog file created

### DIFF
--- a/playbooks/portals-setup-following.yml
+++ b/playbooks/portals-setup-following.yml
@@ -2,18 +2,17 @@
   hosts: webportals
   remote_user: "{{ webportal_user }}"
   gather_facts: True
-  serial: 1              # Limit concurrency
+  serial: 1 # Limit concurrency
   any_errors_fatal: True # Stop on first error, do not execute on the next host
-  vars:                  # Playbook specific vars
+  vars: # Playbook specific vars
     max_hosts: "{{ groups['webportals'] | length - 1 }}"
     lastpass_required: True
     portal_action: "portal-setup"
 
   tasks:
-    
     - name: Check '--limit' is used
       include_tasks: tasks/host-limit-check.yml
-    
+
     - name: Check supported OS and version
       assert:
         that:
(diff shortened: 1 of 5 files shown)


playbooks/portals-setup-following.yml
```diff
@@ -25,19 +24,32 @@
       args:
         apply:
           become: True
-    
-    - name: Include firewall setup
-      include_tasks: tasks/portal-firewall-setup.yml
+
+    # Need to set up firewall then apply ip tables after docker is installed
```
MSevey, line 28: This fails because the ip tables rely on docker which isn't installed until `portal-setup-server.yml` is executed.

playbooks/tasks/portal-role-task-mongo-setup-docker-service.yml
```diff
@@ -84,7 +83,8 @@
         group: "{{ webportal_user }}"
         remove: True
       become: True
-    
+      when: mgkey_stat_result.stat.exists
```
MSevey, line 86: I added this as a stop gap, but there is probably a better fix. We shouldn't be triggering this block at all on new servers when nothing exists. A user shouldn't have to tell ansible to delete the mongo database on a new server set up.

playbooks/tasks/portal-role-task.yml
```diff
@@ -34,8 +33,8 @@
 # Setup MongoDB
 # NOTE: MongoDB docker service must be setup before Sia docker service, because
 # skyd depends on mongo.
-- name: Include ensuring MongoDB node is a member of replicaset
-  include_tasks: tasks/portal-role-task-mongo-replicaset-ensure-member.yml
+# - name: Include ensuring MongoDB node is a member of replicaset
+#   include_tasks: tasks/portal-role-task-mongo-replicaset-ensure-member.yml
```
MSevey, line 37: I had to comment this out for the migration work. It kept failing. Should be able to add it back in.

PR #185, opened 2022-01-05
Create a config.yml

MSevey:
# PULL REQUEST

## Overview
As I work through setting up a portal, I think it makes sense to rename `portal-versions.yml` to `config.yml` as I believe it makes more intuitive sense.

Also, I added some variables that a user might want to update, with this I changed the LastPass folder variables slightly. 

**OUTSTANDING QUESTION** should we extend this to all the `my-vars/file.yml` references in the README? I think for new users, it is more helpful to have a single config.yml file reference that we are setting different variables in versus always changing the name of the file. I think a point of possible confusion is that the name of the file is important. By naming is `config.yml` and keeping that naming throughout the README it would allow for a single line at the top that explains this file can be named whatever the user would like, or left as config. 

## Checklist
 - [ ] Changelog file created


firyx (dismissed):

README.md
```diff
@@ -95,7 +95,7 @@ a playbook.
   - `last-portal-versions.yml` could be used to rerun portal deploy on another
     host (more info below).
 - `my-vars`
-  - content ignored by git (with exception of `portal-versions.sample.do-not-edit.yml`)
+  - content ignored by git (with exception of `config.yml`)
```
ro-tex, line 98:
```suggestion
  - content ignored by git (with exception of `config-sample-do-not-edit.yml`)
```

my-vars/config-sample-do-not-edit.yml
```diff
@@ -0,0 +1,60 @@
...
+# Example
+# webportal_user_authorized_keys: "https://raw.githubusercontent.com/SkynetLabs/skynet-webportal/master/setup-scripts/support/authorized_keys"
+webportal_user_authorized_keys: ""
+
+###############################################################################
+# OPTIONAL
+#
+# Below are a list of variables that are optional depending on the playbook you
+# are running.
+###############################################################################
```
ro-tex, line 54: We sometimes also use `portal_setup_following_skip_on_rerun` and we place it in this file. Let's add it here, unless we prefer it to be added to a different file.

MSevey: I think this file should have mostly mandatory variables and  then the most common optional variables. If we put every single variable in here I think it would get confusing and potentially have unexpected impacts on playbooks if people aren't updating/commenting them out.

firyx: > We sometimes also use `portal_setup_following_skip_on_rerun` and we place it in this file. Let's add it here, unless we prefer it to be added to a different file.

This var was now removed as we do not block IPs during portal setup.

ro-tex: OK, makes sense.

ro-tex (approved):

firyx (approved):

PR #186, opened 2022-01-05
rename email variable to be generic for portal operators

MSevey:
# PULL REQUEST

## Overview
Rename email variable to be generic for portal operators.

### DIFF
--- a/playbooks/group_vars/webportals.yml
+++ b/playbooks/group_vars/webportals.yml
@@ -86,7 +86,7 @@ webportal_git_config:
   - name: user.name
     value: "SkynetLabs Devs"
   - name: user.email
-    value: "{{ skynetlabs_devs_email }}"
+    value: "{{ support_email }}"
 
 # TODO: default in role: False
 webportal_setup_dotfiles_and_dev_tools: True
--- a/playbooks/group_vars/webportals.yml
+++ b/playbooks/group_vars/webportals.yml
@@ -86,7 +86,7 @@ webportal_git_config:
   - name: user.name
     value: "SkynetLabs Devs"
   - name: user.email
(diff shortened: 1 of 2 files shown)


kwypchlo (approved):

firyx:
We should split one email var to `ssl_support_email` and `git_checkout_email` vars, because portal operators could use 2 different emails for this.

playbooks/group_vars/webportals.yml
```diff
@@ -86,7 +86,7 @@ webportal_git_config:
   - name: user.name
     value: "SkynetLabs Devs"
   - name: user.email
-    value: "{{ skynetlabs_devs_email }}"
+    value: "{{ support_email }}"
```
firyx, line 89: We should split one email var to `ssl_support_email` and `git_checkout_email` vars, because portal operators could use 2 different emails for this.

playbooks/templates/.env.j2
```diff
@@ -74,7 +74,7 @@ SKYNET_PORTAL_API="https://{{ cluster_domain }}"
 # Administrator contact email.
 # Needed for communication regarding SSL certification.
 # TODO: Deprecated.
-EMAIL_ADDRESS="{{ skynetlabs_devs_email }}"
+EMAIL_ADDRESS="{{ support_email }}"
```
firyx, line 77: We should split one email var to `ssl_support_email` and `git_checkout_email` vars, because portal operators could use 2 different emails for this.

firyx (changes requested):

MSevey:
@firyx to take over PR

firyx (approved):

ro-tex (approved):

PR #188, opened 2022-01-05
Allow for not blocking IP ranges in setup

MSevey:
# PULL REQUEST

## Overview
Adding the conditional for the IP range blocking variables to not be set so that operators can choose to include this functionality or not.

## Checklist
 - [ ] Changelog file created

### DIFF
--- a/playbooks/tasks/portal-firewall-iptables-block-unblock.yml
+++ b/playbooks/tasks/portal-firewall-iptables-block-unblock.yml
@@ -1,5 +1,4 @@
 ---
-
 # Block and unblock incoming traffic from defined IPs or IP ranges
 
 - name: Load private variables from Ansible private repository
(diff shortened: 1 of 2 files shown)


playbooks/tasks/portal-firewall-iptables-block-unblock.yml
```diff
@@ -20,6 +19,7 @@
     source: "{{ item }}"
     jump: DROP
   loop: "{{ private_vars.incoming_ips_ip_ranges_block }}"
+  when: private_vars.incoming_ips_ip_ranges_block is defined
```
firyx, line 22: I would rather use empty list in private vars then skip the task if undefined.
If the portal operator misconfigures something he will never know this task is skipped.
When we want to skip this task and a couple of other long running tasks during portal setup rerun,
there is already a flag that can be set: `portal_setup_following_skip_on_rerun: True`

playbooks/tasks/portal-firewall-iptables-block-unblock.yml
```diff
@@ -28,3 +28,4 @@
     jump: DROP
     state: absent
   loop: "{{ private_vars.incoming_ips_ip_ranges_unblock }}"
+  when: private_vars.incoming_ips_ip_ranges_unblock is defined
```
firyx, line 31: same

firyx (changes requested):

PR #197, opened 2022-01-12
Accounts limit access

firyx:
# PULL REQUEST

## Overview

Update `.env` generation with `ACCOUNTS_LIMIT_ACCESS`.

## Checklist
 - [x] Changelog file created

### DIFF
--- a/playbooks/templates/.env.j2
+++ b/playbooks/templates/.env.j2
@@ -134,7 +134,9 @@ SKYNET_ACCOUNTS_LOG_LEVEL=info
 {% if webportal_common_config.accounts_limit_access is defined %}
 # Accounts limit access to authenticated users
 # Possible values:
-# - 'authenticated': Only authenticated users can access the portal (default if unset or empty)
+# - 'authenticated': Only authenticated users can access the portal. This is
+#   the default value set by skynet-webportal/docker-compose.accounts.yml if
+#   this variable is empty or undefined.
 # - 'false': All users incl. anonymous users can access the portal (access to portal is public)
 ACCOUNTS_LIMIT_ACCESS={{ webportal_common_config.accounts_limit_access }}
 
--- a/playbooks/templates/.env.j2
+++ b/playbooks/templates/.env.j2
@@ -134,8 +134,8 @@ SKYNET_ACCOUNTS_LOG_LEVEL=info
 {% if webportal_common_config.accounts_limit_access is defined %}
 # Accounts limit access to authenticated users
 # Possible values:
-# - authenticated: Only authenticated users can access the portal (default if this env variable is missing)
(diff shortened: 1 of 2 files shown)


playbooks/templates/.env.j2
```diff
@@ -131,6 +131,14 @@ SKYNET_DASHBOARD_URL=https://account.{{ cluster_domain }}
 # Accounts log level
 SKYNET_ACCOUNTS_LOG_LEVEL=info
 
+{% if webportal_common_config.accounts_limit_access is defined %}
+# Accounts limit access to authenticated users
+# Possible values:
+# - 'authenticated': Only authenticated users can access the portal (default if this env variable is missing)
+# - 'false' or empty: All users incl. anonymous users can access the portal (access to portal is public)
```
ro-tex, line 138: We have this in docker-compose:
```
- ACCOUNTS_LIMIT_ACCESS=${ACCOUNTS_LIMIT_ACCESS:-authenticated} #
```
which I think means that the empty value will be replaced with `authenticated`. I'm not 100% sure about that.

ro-tex (approved):

playbooks/templates/.env.j2
firyx: Oh, you are right, I have read a comment on the variable, not the code.

playbooks/templates/.env.j2
```diff
@@ -131,6 +131,14 @@ SKYNET_DASHBOARD_URL=https://account.{{ cluster_domain }}
 # Accounts log level
 SKYNET_ACCOUNTS_LOG_LEVEL=info
 
+{% if webportal_common_config.accounts_limit_access is defined %}
+# Accounts limit access to authenticated users
+# Possible values:
+# - 'authenticated': Only authenticated users can access the portal (default if unset or empty)
+# - 'false': All users incl. anonymous users can access the portal (access to portal is public)
+ACCOUNTS_LIMIT_ACCESS={{ webportal_common_config.accounts_limit_access }}
```
MSevey, line 139: if the intention is for `authenticated` to be the default, shouldn't we just set it like this:
```
ACCOUNTS_LIMIT_ACCESS={{ webportal_common_config.accounts_limit_access | default('authenticated') }}
```
and leave out the if?

Or is the default behavior of ACCOUNTS_LIMIT_ACCESS being defined somewhere else?
If it is being defined somewhere else, I would extend the comment to say that X system handle the default behavior setting so ansible should not handle the default case, and on needs to respond to if the flag is set or not.

firyx: Default value if empty or unset is handled by `skynet-webportal/docker-compose.accounts.yml`.
The comment was updated.

ro-tex (approved):

MSevey (approved):

PR #198, opened 2022-01-12
fix bug in health-check disable

MSevey:
# PULL REQUEST

## Overview
Noticed a small edge case when disabling the health-check fails due to the container restarting. This shouldn't cause the ansible playbook to fail because the end result is that the health-check isn't running which is the goal.

## Checklist
 - [ ] Changelog file created

### DIFF
--- a/playbooks/tasks/portal-health-check-disable.yml
+++ b/playbooks/tasks/portal-health-check-disable.yml
@@ -12,6 +12,13 @@
   # NOTE: portal_action is defined by the playbook, i.e. for deploys,
   # portal_action='portal-deploy'
   command: 'docker exec health-check cli disable "{{ out_of_lb_message | default(portal_action) }}"'
+  register: docker_health_check_disable_result
+  # Only fail the task if there is an error reported to stderr, unless it
+  # is that the container is restarting, since that means the health check isn't
+  # active already.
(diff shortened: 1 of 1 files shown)


kwypchlo (changes requested):
I don't think it's ok to assume that is health-check is restarting then we're ok to proceed - if the issue was temporary, it will start up and server will be put back to load balancer while we assumed it was out. 

Why was health-check container restarting?

MSevey:
> I don't think it's ok to assume that is health-check is restarting then we're ok to proceed - if the issue was temporary, it will start up and server will be put back to load balancer while we assumed it was out.
> 
> Why was health-check container restarting?

That is valid that it could be a temporary hiccup. I think it in production though, the chances that this happens because of a fluke restart that happened to be timed with an ansible command is pretty small. 

I don't think I have seen a time where a docker service was restarting and corrected itself without intervention on our servers fwiw. 

I ran into this edge case when starting up a portal. The health-check container wasn't starting up along with a number of other containers.  I think this is what we would see the majority of the time. The container is going to be stuck in a restart cycle until the bug is fixed.

firyx (approved):

kwypchlo (approved):

PR #199, opened 2022-01-12
Clean up .env File

MSevey:
# PULL REQUEST

## Overview
Clean up .env file, move all defaults to `playbooks/group_vars/webportals.yml` and only have hardcoded values and blanks in the `.env` file. Update comments to reflect changes.

## Checklist
 - [ ] Changelog file created

### DIFF
--- a/playbooks/group_vars/webportals.yml
+++ b/playbooks/group_vars/webportals.yml
@@ -121,6 +121,11 @@ webportal_server_config_defaults:
 webportal_cluster_config_defaults:
   cookie_hash_key: "{{ lookup('pipe','openssl rand -hex 32') }}"
   cookie_enc_key: "{{ lookup('pipe','openssl rand -hex 32') }}"
+  skynet_db_host: "mongo"
+  skynet_db_pass: "{{ lookup('pipe','openssl rand -base64 32') }}"
+  skynet_db_port: "{{ default_mongo_port }}"
+  skynet_db_replicaset: "skynet"
+  skynet_db_user: "admin"
 
 webportal_common_config_defaults:
   blocker_log_level: "info"
--- a/playbooks/group_vars/webportals.yml
+++ b/playbooks/group_vars/webportals.yml
@@ -113,9 +113,10 @@ webportal_allowance:
 webportal_server_config_defaults:
   domain_name: "{{ ansible_host }}"
   portal_name: "{{ ansible_host }}"
-  portal_modules: ""
+  portal_modules: "m"
   sia_wallet_password: ""
   sia_api_port: 9980
+  sia_api_password: "{{ lookup('pipe','openssl rand -base64 32') }}"
   hsd_api_key: "{{ lookup('pipe','openssl rand -base64 32') }}"
 
 webportal_cluster_config_defaults:
@@ -128,6 +129,8 @@ webportal_cluster_config_defaults:
   skynet_db_user: "admin"
 
 webportal_common_config_defaults:
+  abuse_log_level: "info"
+  accounts_log_level: "info"
   blocker_log_level: "info"
 
 # TODO: move to role default vars
--- a/playbooks/group_vars/webportals.yml
+++ b/playbooks/group_vars/webportals.yml
@@ -113,7 +113,7 @@ webportal_allowance:
 webportal_server_config_defaults:
   domain_name: "{{ ansible_host }}"
   portal_name: "{{ ansible_host }}"
-  portal_modules: "m"
+  portal_modules: ""
   sia_wallet_password: ""
   sia_api_port: 9980
   sia_api_password: "{{ lookup('pipe','openssl rand -base64 32') }}"
--- a/playbooks/templates/.env.j2
+++ b/playbooks/templates/.env.j2
@@ -18,7 +18,9 @@
 # For multi-node cluster these should be different, cluster_domain being the
 # domain accessed publicly and server_domain(s) being address(es) behing a load
 # balancer.
-
+#
+# Defaults set in group_vars/webportals.yml
+#
 ###############################################################################
 # Server specific
 
@@ -56,8 +58,8 @@ PORTAL_NAME={{ webportal_server_config.portal_name }}
 # or
 # PORTAL_MODULES=aj
 #
-# REQUIRED: manually defined by user
-PORTAL_MODULES={{ webportal_server_config.portal_modules }}
+# REQUIRED: manually defined by user, default is mongo
+PORTAL_MODULES={{ webportal_server_config.portal_modules | default('m') }}
 
 # Absolute url to the server API.
 #
@@ -71,8 +73,7 @@ SIA_WALLET_PASSWORD={{ webportal_server_config.sia_wallet_password if (webportal
 
 # The port on which siad is listening, defaults to 9980. 
 #
-# OPTIONAL: manually defined by user, or default will be used
-# NOTE: Default set in group_vars/webportals.yml
+# REQUIRED: manually defined by user, or default will be used
 API_PORT={{ webportal_server_config.sia_api_port }}
 
 # Auto generated secure key for your Handshake service integration.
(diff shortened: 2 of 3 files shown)


playbooks/templates/.env.j2
```diff
@@ -47,42 +55,56 @@ PORTAL_NAME={{ webportal_server_config.portal_name }}
...
 # Sia wallet password
+#
+# REQUIRED: Ansible automatically generated 
 SIA_WALLET_PASSWORD={{ webportal_server_config.sia_wallet_password if (webportal_server_config.sia_wallet_password is defined and webportal_server_config.sia_wallet_password is not none and webportal_server_config.sia_wallet_password != '') else (webportal_server_config.sia_wallet_seed if webportal_server_config.sia_wallet_seed is defined else '') }}
 
-# (optional) The port on which siad is listening, defaults to 9980.
+# The port on which siad is listening, defaults to 9980. 
+#
+# OPTIONAL: manually defined by user, or default will be used
+# TODO: Does this default need to be explicitly set?
```
ro-tex, line 75: API_PORT is only used by `blocker`, if I'm not mistaken, and it doesn't need it to be defined - it will happily use 9980 otherwise.

playbooks/templates/.env.j2
```diff
@@ -107,64 +133,96 @@ AIRTABLE_FIELD={{ webportal_common_config.airtable_field }}
...
 # MongoDB password
+#
+# REQUIRED: Ansible automatically generated 
+# TODO: Ansible doesn't not currently automatically generate this but it should 
 SKYNET_DB_PASS={{ mongodb_config.skynet_db_pass }}
 
 # MongoDB address or container name
+#
+# OPTIONAL: manually defined by user, or default will be used
+# TODO: Do these need to be able to be defined by the user?
```
ro-tex, line 149: I don't think we *need* to make this configurable. Fewer configuration options might actually be better at this stage.

playbooks/templates/.env.j2
```diff
@@ -107,64 +133,96 @@ AIRTABLE_FIELD={{ webportal_common_config.airtable_field }}
...
 SKYNET_DB_HOST=mongo
 
 # MongoDB port
+#
+# OPTIONAL: manually defined by user, or default will be used
+# TODO: Do these need to be able to be defined by the user?
```
ro-tex, line 155: Same as SKYNET_DB_HOST - no need.

playbooks/templates/.env.j2
```diff
@@ -107,64 +133,96 @@ AIRTABLE_FIELD={{ webportal_common_config.airtable_field }}
...
 
 # Absolute url to the (cluster) portal dashboard.
+#
+# REQUIRED: See top of file 
 SKYNET_DASHBOARD_URL=https://account.{{ cluster_domain }}
 
 # Accounts log level
+#
+# OPTIONAL: manually defined by user, or default will be used
+# TODO: Do these need to be able to be defined by the user?
```
ro-tex, line 175: All log levels should be configurable by the user. They are very helpful when debugging an issue. If a portal op reaches out to us on Discord and asks us about an issue, one of the first things I would do is ask them for their logs with `level=trace`, so we can see detailed logging.

playbooks/templates/.env.j2
MSevey: we should just remove this then.

playbooks/templates/.env.j2
MSevey: same question

MSevey:
Made some updates, added some more defaults so most things can be configurable, but don't need to be

kwypchlo (dismissed):

firyx (approved):

playbooks/group_vars/webportals.yml
```diff
@@ -113,16 +113,24 @@ webportal_allowance:
 webportal_server_config_defaults:
   domain_name: "{{ ansible_host }}"
   portal_name: "{{ ansible_host }}"
-  portal_modules: ""
+  portal_modules: "m"
```
kwypchlo, line 116: do we want to also include accounts by default or we want to default to as minimal setup as possible? including just mongo sounds redundant

playbooks/group_vars/webportals.yml
```diff
@@ -113,16 +113,24 @@ webportal_allowance:
...
   sia_wallet_password: ""
   sia_api_port: 9980
+  sia_api_password: "{{ lookup('pipe','openssl rand -base64 32') }}"
```
kwypchlo, line 119: is this gonna work? sia_api_password should equal the one that skyd generates on the first run right?

playbooks/templates/.env.j2
```diff
@@ -138,63 +179,85 @@ SKYNET_ACCOUNTS_LOG_LEVEL=info
...
 ACCOUNTS_LIMIT_ACCESS={{ webportal_common_config.accounts_limit_access }}
 
 {% endif %}
 # Accounts email
+# REQUIRED: manually defined by user
 ACCOUNTS_EMAIL_URI={{ webportal_common_config.accounts_email_uri }}
 
 # Domain to which your cookies will be issued
+#
+# REQUIRED: See top of file
```
kwypchlo, line 192: do we want to note that vars like this are required only if accounts module (or some other module) is enabled?

playbooks/group_vars/webportals.yml
MSevey: yup, skyd checks for an environment variable first. So even if it previously created one on disk, when it restarts it will start using the new env var.

playbooks/group_vars/webportals.yml
MSevey: I went with the minimum, so no accounts. Since our docker compose files set the mongo env vars for skyd, skyd won't start without mongo.  So even though mongo isn't required for a bare bones skyd instance, because ansible is using our docker compose files, mongo becomes required. 

tl;dr minimum set up requires mongo.

playbooks/templates/.env.j2
MSevey: this is already grouped like that :-)

This is under the `(optional) Accounts` section.

firyx (approved):

playbooks/group_vars/webportals.yml
kwypchlo: > Since our docker compose files set the mongo env vars for skyd, skyd won't start without mongo

This is not correct any more, I fixed that couple of days ago when I tried starting without mongo and it worked fine :) https://github.com/SkynetLabs/skynet-webportal/commit/dba33f833569968b8656ca8b53a3cec5d7090caf

playbooks/group_vars/webportals.yml
kwypchlo: This will break nginx reading api password since it expects it to be stored on a disk in our setup.

https://github.com/SkynetLabs/skynet-webportal/blob/master/docker/nginx/libs/skynet/utils.lua

playbooks/templates/.env.j2
```diff
@@ -95,41 +118,59 @@ DISCORD_MENTION_USER_ID={{ discord_mention_user_id | default('') }}
...
 # MongoDB password
+#
+# REQUIRED: manually defined by user, or Ansible automatically generated 
 SKYNET_DB_PASS={{ mongodb_config.skynet_db_pass }}
 
 # MongoDB address or container name
-SKYNET_DB_HOST=mongo
+#
+# REQUIRED: manually defined by user, or default will be used
+SKYNET_DB_HOST={{ mongodb_config.skynet_db_host }}
```
firyx, line 149: Is this configurable for all services using mongo?
Isn't `mongo` container service/host name hardcoded in some docker services using mongo?

firyx (changes requested):

playbooks/group_vars/webportals.yml
MSevey: Plan is to update nginx

playbooks/templates/.env.j2
MSevey: There are two hardcoded references 

https://github.com/SkynetLabs/skynet-webportal/blob/master/docker/malware-scanner/Dockerfile#L16
https://github.com/SkynetLabs/skynet-webportal/blob/master/docker/accounts/Dockerfile#L16

But then they are overwritten in the docker-compose files

https://github.com/SkynetLabs/skynet-webportal/blob/master/docker-compose.accounts.yml#L40

So since nothing is hardcoded in the docker compose files, this means that they are configurable right?

playbooks/group_vars/webportals.yml
MSevey: Nice 👍 

I think we can leave this for now. If we want to change it in the future we can.

playbooks/group_vars/webportals.yml
kwypchlo: resolved in https://github.com/SkynetLabs/skynet-webportal/pull/1567

playbooks/group_vars/webportals.yml
MSevey, line 116:
```suggestion
  portal_modules: ""
```

firyx (approved):

kwypchlo (approved):

PR #205, opened 2022-01-20
remove not configurable variables

kwypchlo:
BLOCKER_HOST and BLOCKER_PORT are not user configurable and should just be configured in docker-compose file

requires https://github.com/SkynetLabs/skynet-webportal/pull/1573


MSevey (approved):

peterjan (approved):
reminder that we should only merge after https://github.com/SkynetLabs/skynet-webportal/pull/1573 is merged

peterjan:
it was set to auto-merge apparently, kind of dangerous if it's depending on another PR
I'll ping Karol to ensure we don't run into troublehere, should be fine though since it's
only a playbook.

PR #212, opened 2022-02-07
Change the S3 backup dir from per-server to one common dir.

ro-tex:
# PULL REQUEST

## Overview

Since our DB is cluster-wide, we don't need to backup each server individually. Moreover, the backup and restore scripts were not written to work with individual backups. That's why I'm changing the default path to be a common `mongo` dir under the S3 bucket.

PS: Sorry for the changed whitespaces - my editors are cleaning those up automatically, so I would have needed to add them back manually in order to reduce the size of the PR.

## Checklist
 - [ ] Changelog file created

### DIFF
--- a/playbooks/templates/.env.j2
+++ b/playbooks/templates/.env.j2
@@ -28,24 +28,24 @@
 # This is the domain that will be used for generating certificates and nginx
 # routing for cluster server together with PORTAL_DOMAIN.
 #
-# REQUIRED if different from PORTAL_DOMAIN: See top of file 
+# REQUIRED if different from PORTAL_DOMAIN: See top of file
 SERVER_DOMAIN={{ server_domain }}
 
 {% endif %}
 # A string representing name of your portal e.g. 'my-portal.com' or 'My Awesome
 # Skynet Portal' (internal use only)
 #
-# REQUIRED: Manually defined by user, or Ansible automatically generated 
+# REQUIRED: Manually defined by user, or Ansible automatically generated
 PORTAL_NAME={{ webportal_server_config.portal_name }}
 
 # Server UID to uniquelly identify the server. Used by Blocker module.
 #
-# REQUIRED: Manually defined by user, or Ansible automatically generated 
+# REQUIRED: Manually defined by user, or Ansible automatically generated
 SERVER_UID={{ webportal_server_config.server_uid }}
 
 # Sia api password.
 #
-# REQUIRED: Manually defined by user, or Ansible automatically generated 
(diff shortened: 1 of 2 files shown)


playbooks/templates/.env.j2
```diff
@@ -120,7 +120,7 @@ DISCORD_MENTION_USER_ID={{ discord_mention_user_id | default('') }}
 # For dev servers: {{ webportal_common_config.s3_backup_path }}_<server hostname>
```
firyx, line 120: With this change these 3 comments
- `# common path for...`
- `# For prod servers...`
- `# For dev servers...`
are not relevant anymore and should be removed.

firyx (changes requested):

playbooks/templates/.env.j2
```diff
@@ -120,7 +120,7 @@ DISCORD_MENTION_USER_ID={{ discord_mention_user_id | default('') }}
...
 #
 # OPTIONAL: manually defined by user, can be blank
-S3_BACKUP_PATH={{ webportal_common_config.s3_backup_path }}{{ '_' + inventory_hostname if 'webportals_prod' not in group_names else '' }}
+S3_BACKUP_PATH={{ webportal_common_config.s3_backup_path }}/mongo
```
firyx, line 123: `webportal_common_config.s3_backup_path` is for us currently set in `common.yml` in LastPass, which means with the change it will be shared between all `prod`, `dev`, `dev-cc` clusters. That's not ok, right?

We should either make it cluster specific like:  
`S3_BACKUP_PATH={{ webportal_common_config.s3_backup_path }}/{{ portal_cluster_id }}-mongo` (I'm inclined to this)

or move this setting from `common.yml` to cluster yamls and make them manually unique between clusters.

And update `.env` vars on all servers in both cases.

firyx (changes requested):

playbooks/templates/.env.j2
ro-tex: I also like the `S3_BACKUP_PATH={{ webportal_common_config.s3_backup_path }}/{{ portal_cluster_id }}-mongo` approach. In that case we might actually drop the `-mongo` suffix. The env var is defined as a generic backup path, so it should be reusable.

I'll make the changes here.

playbooks/templates/.env.j2
ro-tex: Removed.

firyx (approved):

MSevey (approved):

PR #217, opened 2022-02-08
do not prune current docker data

kwypchlo:
- completely drop pruning docker data **before** rebuilding containers - we want to leverage build cache to speed up the process
- **after** rebuilding, prune only docker data that is more than 4 hours old (arbitrary number but should be good enough) which means it will keep docker data that was produced during current build process

Originally we added strict pruning before and after because we were running out of disk space during deploys due to issues with cache, small disks on some machines, log files overflow etc. Also with every rebuild, docker data kept growing and I didn't know there was an option to prune just old data. I think we're in a quite good position to keep just current docker data now and it should significantly improve build times on deploys.

Let's test it @firyx :)

### DIFF
--- a/playbooks/tasks/portal-docker-prune-old-data.yml
+++ b/playbooks/tasks/portal-docker-prune-old-data.yml
@@ -1,0 +1,11 @@
+---
+
+# Prune old docker data
+
+# TODO: Use community.docker.docker_prune with long timeout (requires
+# community.docker upgrade)
+# Prune old docker data (do not prune data produced in current build process)
+- name: Prune old docker data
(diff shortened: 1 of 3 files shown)


MSevey (changes requested):
Now that it is just after, we can move it to a task to DRY up the code.

firyx (approved):

firyx (approved):

kwypchlo:
> Now that it is just after, we can move it to a task to DRY up the code.

@firyx added that in https://github.com/SkynetLabs/ansible-playbooks/pull/217/commits/bc3b89d42b6f619595741d8ad10b0946c1c73ff7

MSevey (approved):

PR #220, opened 2022-02-10
Set default values

firyx:
# PULL REQUEST

## Overview

Portal setup scripts used to setup additional toools or settings on portal servers according to Skynet Labs needs. Like:
- ufw firewall
- blocking outgoing traffic to local networks
- timezone
- dotfiles
- dev tools
- ...
These defaults might not be desired by community portal operators, they might want to run alternative tools or they might want to run other services than just skynet webportal stack.

This PR makes all these settings optional, turned off by default.

## Checklist
 - [ ] Changelog file created

### DIFF
--- a/my-vars/config-sample-do-not-edit.yml
+++ b/my-vars/config-sample-do-not-edit.yml
@@ -1,5 +1,6 @@
 # This config file can be used to set any ansible variables needed to execute
-# necessary ansible playbooks.
+# necessary ansible playbooks. These settings when used will also override
+# values defined in playbooks/group_vars/webportals.yml.
 
 ###############################################################################
 # REQUIRED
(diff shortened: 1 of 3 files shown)


firyx:
First commit is just a sample draft of the changes, there are more changes coming.

my-vars/config-sample-do-not-edit.yml
```diff
@@ -8,6 +9,32 @@
 # suitable.
 ###############################################################################
 
+###############################################################################
+# Server Setup
```
MSevey, line 13: I would add something like
```
below are some of the recommended server level settings for portals. If you wish to disable any of these settings, uncomment the variable and change the value. 
PLEASE NOTE: we have not tested any non-default conditions for these settings, so you should be comfortable debugging the server and portal on your own if you plan to change these settings. 
```

my-vars/config-sample-do-not-edit.yml
```diff
@@ -8,6 +9,32 @@
...
+#
+# Ansible script to setup and later to update portal server
```
MSevey, line 15: This comment doesn't make sense. I'm not sure what we are trying to convey to the portal operator here.

my-vars/config-sample-do-not-edit.yml
firyx: Added.

my-vars/config-sample-do-not-edit.yml
firyx: Replaced with comments above.

ro-tex (approved):

MSevey (approved):

PR #233, opened 2022-02-25
remove docker compose orphaned services on deploy

kwypchlo:
adds `--remove-orphans` on deploy, useful when:
- we changed container name of one service and want to clean up the old service on restart
- we changed PORTAL_MODULES and want to clean up services not needed any more on the server

### DIFF
--- a/playbooks/portals-takedown.yml
+++ b/playbooks/portals-takedown.yml
@@ -47,6 +47,7 @@
         project_src: "{{ webportal_dir }}"
         files: "{{ webportal_docker_compose_files_wanted }}"
         build: False
+        remove_orphans: True
         nocache: True
         pull: True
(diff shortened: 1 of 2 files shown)


ro-tex (approved):

playbooks/portals-takedown.yml
```diff
@@ -47,6 +47,7 @@
         project_src: "{{ webportal_dir }}"
         files: "{{ webportal_docker_compose_files_wanted }}"
         build: False
+        remove_orphans: True
```
peterjan, line 50: this is probably fine but is it really necessary? This might erase container data in times where we bring down a portal for debugging purposes, is that right?

peterjan (approved):

PR #234, opened 2022-02-25
Add handling for different path separators for lastpass

MSevey:
# PULL REQUEST

## Overview
This PR adds handling of the different path separators for LastPass directories between shared folders and personal folders. Since shared folders have the prefix `Shared`, we can easily set the separator based on the top level folder name.

## Example for Visual Changes

Terminal Output proof:

https://hackmd.io/so0iuxfgQxe2mQwhasrB8Q

## Checklist
 - [x] Changelog file created

### DIFF
--- a/changelog/items/bugs-fixed/personal-laspass-handling.md
+++ b/changelog/items/bugs-fixed/personal-laspass-handling.md
@@ -1,0 +1,1 @@
+- Add handling for personal lastpass accounts with the different path separator for folders.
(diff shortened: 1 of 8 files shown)


firyx (changes requested):
When I was testing personal LastPass, just changing the separator was not enough.

What didn't seem to work correctly was at least a task checking if the LastPass record already exists. IIRC `portal-setup-following` was always updating `user` user password in LastPass, regardless it was already set, we lost control over the server and had to reset its OS. I didn't had time to investigate further, but I think the issue could be fixed by quoting correctly all ansible tasks using `lpass` and the record(s) path.

PR #235, opened 2022-02-28
Expand set-allowance-max-storage-price 

ro-tex:
# PULL REQUEST

## Overview

**Problem:** 
`max-contract-price` and `max-sector-access-price` are only set on server creation and cannot be adjusted via ansible scripts after that.

**Solution:**
Expand `set-allowance-max-storage-price` to also set `max-contract-price` and `max-sector-access-price`.
If this is not the way we want to approach this, we can always close this PR.

## Checklist
 - [ ] Changelog file created

### DIFF
--- a/playbooks/portals-set-allowance-price-controls.yml
+++ b/playbooks/portals-set-allowance-price-controls.yml
@@ -29,3 +29,6 @@
       ansible.builtin.command: |
         docker exec sia siac renter setallowance --max-storage-price {{ max_storage_price }} --max-contract-price {{ max_contract_price }} --max-sector-access-price {{ max_sector_access_price }}
       register: set_allowance_result
+      # Verify result to prevent mistakes and misconfiguration, because
+      # setallowance might not fail but just switch to interactive mode.
+      failed_when: set_allowance_result.stdout.find("Allowance updated.") == -1
--- a/playbooks/portals-set-allowance-price-controls.yml
+++ b/playbooks/portals-set-allowance-price-controls.yml
@@ -26,4 +26,11 @@
 
     # Set allowance
     - name: Set allowance
-      command: "docker exec sia siac renter setallowance --max-storage-price {{ max_storage_price }} --max-contract-price {{ max_contract_price }} --max-sector-access-price {{ max_sector_access_price }}"
+      ansible.builtin.command: |
+        docker exec sia siac renter setallowance --max-storage-price {{ max_storage_price }} --max-contract-price {{ max_contract_price }} --max-sector-access-price {{ max_sector_access_price }}
+      register: set_allowance_result
+
+    docker exec sia siac renter setallowance
+      --max-contract-price {{ max_contract_price }}
+      --max-sector-access-price {{ max_sector_access_price }}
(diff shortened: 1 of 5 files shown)


firyx (dismissed):

MSevey:
GitHub is annoying that it doesn't let you comment on code outside the updates.....

I think we should update this script in a few ways. 

- [ ] We should remove this hardcoded value https://github.com/SkynetLabs/ansible-playbooks/pull/235/files#diff-3707d920c6e46112290d089357cdcfd132eb695d19879fe52de3cb9283525958R15
- [ ] We should rename it, `portals-set-allowance-price-controls.yml`
- [ ] we should standardize the ansible command with `ansible.builtin.command` like so

```
  ansible.builtin.command: >
    docker exec sia siac renter setallowance
      --max-contract-price {{ max_contract_price }}
      --max-sector-access-price {{ max_sector_access_price }}
      --max-storage-price {{ max_storage_price }}
```

ro-tex:
> we should standardize the ansible command with ansible.builtin.command like so

I'm not completely sure what you mean by that but I looked through the repo and it seems like `Download skynet-js tags via Github API` is doing a similar thing, so I copied the approach here. Hopefully, that's close to what you had in mind. Otherwise, we can always discuss on discord or you can make the changes you want to see, in case that's easier than explaining it.

playbooks/portals-set-allowance-price-controls.yml
```diff
@@ -20,10 +20,12 @@
...
     - name: Include preparing portal prerequisities
       include_tasks: tasks/portals-prepare.yml
 
     # Set allowance
     - name: Set allowance
-      command: "docker exec sia siac renter setallowance --max-storage-price {{ max_storage_price }}"
\ No newline at end of file
+      ansible.builtin.command: |
+        docker exec sia siac renter setallowance --max-storage-price {{ max_storage_price }} --max-contract-price {{ max_contract_price }} --max-sector-access-price {{ max_sector_access_price }}
+      register: set_allowance_result
```
ro-tex, line 31: This is not used anywhere. Do we need it?

MSevey:
> > we should standardize the ansible command with ansible.builtin.command like so
> 
> I'm not completely sure what you mean by that but I looked through the repo and it seems like `Download skynet-js tags via Github API` is doing a similar thing, so I copied the approach here. Hopefully, that's close to what you had in mind. Otherwise, we can always discuss on discord or you can make the changes you want to see, in case that's easier than explaining it.

I gave you a copy paste example :-)

playbooks/portals-set-allowance-price-controls.yml
MSevey: well....you added it just now :-) 
So if you aren't using it then you can delete it.

ro-tex: I did some "creative borrowing" from Fil's code without fully understanding it. ;)
I'll just double down on that and add the failure check he has.

firyx (approved):

MSevey (approved):

PR #240, opened 2022-03-07
Tor blocklist cron

firyx:
# PULL REQUEST

## Overview

This PR updates a playbook to block Tor exit nodes traffic and updates `portal-setup-following` to use new Tor traffic blocking task. Default option is to block Tor traffic, operators can opt out.

## Checklist
 - [x] Changelog file created

### DIFF
--- a/playbooks/group_vars/webportals.yml
+++ b/playbooks/group_vars/webportals.yml
@@ -49,6 +49,23 @@ local_networks_denied:
   - "198.18.0.0/15"
   - "169.254.0.0/16"
 
+# Iptables: Tor exit nodes blocking
+block_tor_exit_nodes: True
+block_tor_exit_nodes_lists:
+  torproject:
+    url: "https://check.torproject.org/torbulkexitlist?ip="
+    filename: "torproject-exitnodelist"
+    ipset_ipv4: "torproject-exitnodelist-ipv4"
+  dan:
+    url: "https://www.dan.me.uk/torlist/?exit"
+    filename: "dan-exitnodelist"
+    ipset_ipv4: "dan-exitnodelist-ipv4"
+    ipset_ipv6: "dan-exitnodelist-ipv6"
+  # Downloads from dan.me.uk are limited to once per 30 mins (= 1800 secs), we
+  # do not redownload the given number of seconds to prevent errors loading
+  # error message instead of IP list to ipset
+  do_not_redownload_within_secs: 2000
+
 # LastPass settings
 lastpass_ansible_dir: "Shared-Ansible"
 
--- a/playbooks/templates/tor-blocklists-update.sh.j2
+++ b/playbooks/templates/tor-blocklists-update.sh.j2
@@ -61,5 +61,5 @@ if [[ ! -f $timestamp_file || $(cat $timestamp_file) -lt $past ]]; then
     echo "Creating/Updating dan timestamp"
     echo "$(date +%s)" > $timestamp_file
 else
-    echo "It's too soon to redownload dan exitnodelist"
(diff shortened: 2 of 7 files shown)


playbooks/group_vars/webportals.yml
```diff
@@ -49,6 +49,23 @@ local_networks_denied:
...
+    ipset_ipv4: "torproject-exitnodelist-ipv4"
+  dan:
+    url: "https://www.dan.me.uk/torlist/?exit"
+    filename: "dan-exitnodelist"
+    ipset_ipv4: "dan-exitnodelist-ipv4"
+    ipset_ipv6: "dan-exitnodelist-ipv6"
+  # Downloads from dan.me.uk are limited to once per 30 mins (= 1800 secs), we
+  # do not redownload the given number of seconds to prevent errors loading
+  # error message instead of IP list to ipset
+  do_not_redownload_within_secs: 2000
```
ro-tex, line 67: nit: It would be nice to add a comment here explaining that we're caching the result in `/tmp` in order to avoid redownloading it.

playbooks/templates/tor-blocklists-update.sh.j2
```diff
@@ -0,0 +1,65 @@
...
+    echo "Downloading torproject exitnodelist"
+    curl -sSL $url -o $file
+
+    echo "Loading torproject exitnodelist to ipset"
+    cat $file | xargs -n 1 ipset -exist add $ipset 
+
+    echo "Creating/Updating torproject timestamp"
+    echo "$(date +%s)" > $timestamp_file
+else
+    echo "It's too soon to redownload torproject exitnodelist"
```
ro-tex, line 30: This might misinform users that the tor list is not being used. How about something like this?

```suggestion
    echo "It's too soon to redownload torproject exitnodelist, using cached version from $(cat $timestamp_file)."
```

playbooks/templates/tor-blocklists-update.sh.j2
```diff
@@ -0,0 +1,65 @@
...
+    cat $file | grep -E "^[0-9]+\.[0-9]+\.[0-9]+\.[0-9]+$" | xargs -n 1 ipset -exist add $ipset_ipv4
+
+    echo "Loading dan exitnodelist to ipset IPv6"
+    # grep only IPv6
+    cat $file | grep -E ":" | xargs -n 1 ipset -exist add $ipset_ipv6
+
+    echo "Creating/Updating dan timestamp"
+    echo "$(date +%s)" > $timestamp_file
+else
+    echo "It's too soon to redownload dan exitnodelist"
```
ro-tex, line 64: Same as above:
```suggestion
    echo "It's too soon to redownload dan exitnodelist, using cached version from $(cat $timestamp_file)."
```

ro-tex (approved):
LGTM.

Added minor suggestions.

playbooks/group_vars/webportals.yml
firyx: The redownloads are handled by the shell script,
so user or cronjob can run the script without any restrictions (like 3x in 1 minute or every 5 mins)

During download a timestamp file is saved, if we download every 5 mins, the script will skip downloading until (timestamp in a file + `do_not_redownload_within_secs`) < now time.

ro-tex: Yes, I noticed that. What I mean by this comment is that it would be nice to explain that here with a single sentence, so people won't wonder what happens and go to read the shell script in order to find out.

ro-tex (approved):

MSevey (approved):

PR #246, opened 2022-03-09
Move cron script to devops folder to avoid path errors

MSevey:
# PULL REQUEST

## Overview
The current `portals-setup-following` script would throw errors due to the updates to the cronjob handling. This addresses those errors by updating the folder that they are downloaded to during setup. 

## Checklist
 - [ ] Changelog file created

### DIFF
--- a/playbooks/tasks/portal-firewall-iptables-block-tor-exit-nodes.yml
+++ b/playbooks/tasks/portal-firewall-iptables-block-tor-exit-nodes.yml
@@ -1,5 +1,4 @@
 ---
-
 # Block Tor exit node traffic
 
 # Requirements:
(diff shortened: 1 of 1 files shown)


firyx (approved):
Before we deploy this to all portals, we should test on 1 host that `ansible.builtin.cron` task updates existing cron job with new script path (vs. creating a second cron job).

playbooks/tasks/portal-firewall-iptables-block-tor-exit-nodes.yml
```diff
@@ -36,30 +35,30 @@
...
   ansible.builtin.file:
-    path: "{{ webportal_dir }}/scripts"
+    path: "{{ devops_scripts_dir }}"
     state: directory
 
 - name: Upload script to update tor exit node lists (used by root cron job)
   ansible.builtin.template:
     src: templates/tor-blocklists-update.sh.j2
-    dest: "{{ webportal_dir }}/scripts/tor-blocklists-update.sh"
+    dest: "{{ devops_scripts_dir }}/tor-blocklists-update.sh"
```
firyx, line 44: nit: looking at this now, we might want to create another var for script path in `webportals.yml` something like: `block_tor_exit_nodes_script_file: "{{ devops_scripts_dir }}/tor-blocklists-update.sh"`

firyx (approved):

ro-tex (approved):

playbooks/tasks/portal-firewall-iptables-block-tor-exit-nodes.yml
MSevey: Need to run this test still

run this against any server, no need for sudo/become/lastpass
```
- name: Test cron job idempotency
  ansible.builtin.cron:
    name: "test 1"
    minute: "45"
    job: "echo abc"
```

then on the server crontab -l to check the cronjob is present
then update a la updating the "script path"
```
- name: Test cron job idempotency
  ansible.builtin.cron:
    name: "test 1"
    minute: "45"
    job: "echo def"
```

then crontab -l => we should have just 1 our cron job with echo def (echo abc should be overriden)

then remove the test job either via crontab -e or
```
- name: Test cron job idempotency
  ansible.builtin.cron:
    name: "test 1"
    state: absent
```

MSevey: run and confirmed

PR #250, opened 2022-03-09
Switch ignore_errors to failed_when in setup following

MSevey:
# PULL REQUEST

## Overview
In `portals-setup-following` we have code to gracefully shut down docker if it is running. This is so that we can rerun the script on portals to update settings. 

We used `ignore_errors: True` on two tasks that could fail if docker isn't set up, i.e. a new portal. This allowed us to check the results in subsequent tasks to determine if we needed to do anything. 

`ignore_errors` still prints out the error in the ansible logs, so users were getting alarmed and confused when they would see the red errors from these tasks when setting up their portals. 

By switching to `failed_when: False` we achieve the same result but with no red error log. This just required updating the following task to look at the errors from these tasks rather than the failed result. 

## Checklist
 - [x] Changelog file created

### DIFF
--- a/playbooks/portals-setup-following.yml
+++ b/playbooks/portals-setup-following.yml
@@ -39,16 +39,28 @@
     - name: Check docker was setup
       ansible.builtin.command: docker ps
       register: docker_ps_result
-      ignore_errors: True
+      # Use failed_when: False instead of ignore_errors: True so that here isn't
+      # a red logs that are confusing to users.
+      #
+      # NOTE: failed_when: False forces docker_ps_result.failed to be False but
+      # any errors are still recording in docker_ps_result.stderr
+      failed_when: False
 
     - name: Check docker SDK for Python3 was installed
       ansible.builtin.command: python3 -m pip show docker
       register: docker_pip_result
-      ignore_errors: True
+      # Use failed_when: False instead of ignore_errors: True so that here isn't
+      # a red logs that are confusing to users.
+      #
+      # NOTE: failed_when: False forces docker_pip_result.failed to be False but
+      # any errors are still recording in docker_pip_result.stderr
+      failed_when: False
 
+    # If both of the previous tasks were successful and didn't return any
+    # errors, we want to try and stop any active docker containers.
     - name: Include stopping portal and other docker containers (if exist)
       include_tasks: tasks/portal-stop-and-docker-containers-stop.yml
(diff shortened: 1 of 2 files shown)


ro-tex (dismissed):

playbooks/portals-setup-following.yml
```diff
@@ -39,16 +39,28 @@
...
+      # a red logs that are confusing to users.
+      #
+      # NOTE: failed_when: False forces docker_ps_result.failed to be False but
+      # any errors are still recording in docker_ps_result.stderr
+      failed_when: False
 
     - name: Check docker SDK for Python3 was installed
       ansible.builtin.command: python3 -m pip show docker
       register: docker_pip_result
-      ignore_errors: True
```
firyx, line 47: Here you need to handle an error when pip or docker is not yet installed:
`failed_when: False`

playbooks/portals-setup-following.yml
```diff
@@ -39,16 +39,28 @@
...
+      # NOTE: failed_when: False forces docker_pip_result.failed to be False but
+      # any errors are still recording in docker_pip_result.stderr
+      failed_when: False
+
+    # If both of the previous tasks were successful and didn't return any
+    # errors, we want to try and stop any active docker containers.
     - name: Include stopping portal and other docker containers (if exist)
       include_tasks: tasks/portal-stop-and-docker-containers-stop.yml
-      when: ((docker_ps_result.failed | default(False)) or (docker_pip_result.failed | default(False))) != True
+      when: docker_ps_result.stderr == '' and docker_pip_result.stderr == ''
```
firyx, line 63: It's safer to use exit code, because stderr could be empty, but command(s) might be failing.

https://docs.ansible.com/ansible/latest/collections/ansible/builtin/command_module.html#return-rc

firyx (changes requested):

playbooks/portals-setup-following.yml
MSevey: Are you saying I need `failed_when: False`? Because I have it. L57

firyx: I have missed that, then it's ok.

firyx (changes requested):

ro-tex (approved):

firyx (approved):

PR #253, opened 2022-03-10
Fix Mongo Init for Single Server

MSevey:
# PULL REQUEST

## Overview
This PR fixes the mongo initialization for a single server. 

Tested on dev3 as well to ensure it doesn't impact existing mongo cluster. 

## Checklist
 - [x] Changelog file created

### DIFF
--- a/playbooks/tasks/portal-role-task-mongo-setup-docker-service.yml
+++ b/playbooks/tasks/portal-role-task-mongo-setup-docker-service.yml
@@ -1,5 +1,4 @@
 ---
-
 # Ensure MongoDB docker service is setup correctly
 # TODO: This ansible playbook task, should be moved to portal role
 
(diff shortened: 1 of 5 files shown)


playbooks/tasks/portal-role-task-mongo-setup-docker-service.yml
```diff
@@ -25,6 +24,8 @@
   when: mgkey_stat_result.stat.isreg | default(False)
   register: mgkey_update_result
```
ro-tex, line 25: Not relevant to this PR but in here we might want to ensure the proper ownership and access rights of `mgkey` - it needs to be owned by user with id 999, which corresponds to `mongo` in the official MongoDB docker container. It also needs to have its access rights set to 0400.

ro-tex (approved):

playbooks/tasks/portal-role-task-mongo-setup-docker-service.yml
MSevey: probably worth a follow up.

MSevey: if we check the permissions, and they are wrong, we can skip all the other checks and move to resetting the server since with the wrong permissions, mongo won't start.

MSevey:
pending Fil's mongo fixes

MSevey:
obselete

PR #254, opened 2022-03-11
Add option to include deployment during portal setup

firyx:
# PULL REQUEST

## Overview

This PR adds an option to perform portal deployment at the end of `portals-setup-following`.

It is controlled by an optional `deploy_after_setup` variable.

By default (if the above variable is missing) deployment is not performed.

Also note that `parallel_deploys` variable was renamed to `parallel_executions`

## Checklist
 - [x] Changelog file created

### DIFF
--- a/README.md
+++ b/README.md
@@ -541,8 +541,7 @@ To run:
 
 ### Setup Portal from Scratch
 
-Setup process requires 3 playbooks (or 2 playbooks if you enable portal deploy
-during portal setup, see playbook `portal-setup-following` below):
+Setup process requires 3 playbooks:
 
 - `portals-setup-initial.sh` (run once)
 - `portals-setup-following.sh`
--- a/README.md
+++ b/README.md
@@ -221,7 +221,7 @@ or:
 
 By default portals-deploy playbook performs deployments one server at a time
 (rolling updates/deploys). You can enable parallel deployments (deploy to the
-given number of hosts in parallel) by setting optional `parallel_deploys`
+given number of hosts in parallel) by setting optional `parallel_executions`
 variable in used `config.yml`.
 
 Example `config.yml`:
@@ -232,7 +232,7 @@ portal_repo_version: "deploy-2021-08-24"
 portal_skyd_version: "deploy-2021-08-24"
 portal_accounts_version: "deploy-2021-08-23"
 
-parallel_deploys: 3
+parallel_executions: 3
 ```
 
 #### How to Set Deploy Batch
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -541,7 +541,8 @@ To run:
 
 ### Setup Portal from Scratch
 
-Setup process requires 3 playbooks:
+Setup process requires 3 playbooks (or 2 playbooks if you enable portal deploy
+during portal setup, see playbook `portal-setup-following` below):
```
MSevey, line 545: I would revert this.

New portals can't run the deploy as part of setup following in their current state because of the consensus needing to sync.

MSevey (changes requested):

README.md
firyx: done

kwypchlo (commented):
let's make it default to deploy after setup-following - I don't see a value in running setup-following without a deploy but correct me if I'm wrong

MSevey:
> let's make it default to deploy after setup-following - I don't see a value in running setup-following without a deploy but correct me if I'm wrong

@kwypchlo the reason we don't want that as the default is because then the default behavior when setting up a portal will fail. 

The default behavior for setup following should be to complete the setup of a new portal, where the end state is a server running with the skyd instance syncing.  Since the skyd instance isn't synced all tests will fail for a multitude of reasons.

MSevey (approved):

kwypchlo:
> > let's make it default to deploy after setup-following - I don't see a value in running setup-following without a deploy but correct me if I'm wrong
> 
> @kwypchlo the reason we don't want that as the default is because then the default behavior when setting up a portal will fail.
> 
> The default behavior for setup following should be to complete the setup of a new portal, where the end state is a server running with the skyd instance syncing. Since the skyd instance isn't synced all tests will fail for a multitude of reasons.

In that case I would remove option to deploy from [portals-setup-following.yml](https://github.com/SkynetLabs/ansible-playbooks/blob/master/playbooks/portals-setup-following.yml) but add the option run setup-following in [portals-deploy.yml](https://github.com/SkynetLabs/ansible-playbooks/blob/master/playbooks/portals-deploy.yml) - it makes more sense since we wouldn't have to think whether we need to run one script or the other when we deploy.

Note: not sure if we want to run whole setup-following, in the past we needed only 4 specific things in general:
- rebuild ufw with new ips blocked
- recreate .env file with new variables
- recreate crontab
- recreate auth keys file

kwypchlo (approved):
I'm not super happy with the decision to use setup-following with the deploy but technically the code is correct so I'm approving

PR #263, opened 2022-03-21
Update blocker config

peterjan:
# PULL REQUEST

## Overview
This PR adds a dynamic variable to update the env variable `BLOCKER_PORTALS_SYNC`.
It also cleans up some unused variables.

## Example for Visual Changes
N/A

## Checklist
 - [ ] Changelog file created

## Issues Closed
N/A

### DIFF
--- a/playbooks/templates/.env.j2
+++ b/playbooks/templates/.env.j2
@@ -233,6 +233,9 @@ BLOCKER_LOG_LEVEL={{ webportal_common_config.blocker_log_level }}
 BLOCKER_PORTALS_SYNC={{ webportal_common_config.blocker_portals_sync |
 default('')}}
 
+# The following environment variables have been removed from the blocker but are
+# kept here to preserve backwards compatibility.
+#
 # OPTIONAL: manually defined by user
 BLOCKER_NGINX_CACHE_PURGE_LOCK={{
 webportal_common_config.blocker_nginx_cache_purge_lock | default('') }}
--- a/playbooks/templates/.env.j2
+++ b/playbooks/templates/.env.j2
@@ -237,10 +237,8 @@ default('')}}
 # kept here to preserve backwards compatibility.
 #
 # OPTIONAL: manually defined by user
-BLOCKER_NGINX_CACHE_PURGE_LOCK={{
-webportal_common_config.blocker_nginx_cache_purge_lock | default('') }}
-BLOCKER_NGINX_CACHE_PURGE_LIST={{
-webportal_common_config.blocker_nginx_cache_purge_list | default('') }}
+BLOCKER_NGINX_CACHE_PURGE_LOCK={{ webportal_common_config.blocker_nginx_cache_purge_lock | default('') }}
+BLOCKER_NGINX_CACHE_PURGE_LIST={{ webportal_common_config.blocker_nginx_cache_purge_list | default('') }}
 
 {% endif %}
 {% if portal_abuse_scanner_on %}
--- a/playbooks/templates/.env.j2
+++ b/playbooks/templates/.env.j2
@@ -227,9 +227,8 @@ SKYNET_ACCOUNTS_PORT=3000
 # REQUIRED: manually defined by user, or default will be used
 BLOCKER_LOG_LEVEL={{ webportal_common_config.blocker_log_level }}
 
-# OPTIONAL: manually defined by user, can be blank
-BLOCKER_NGINX_CACHE_PURGE_LIST={{ webportal_common_config.blocker_nginx_cache_purge_list | default('') }}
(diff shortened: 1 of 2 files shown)


playbooks/templates/.env.j2
```diff
@@ -227,9 +227,8 @@ SKYNET_ACCOUNTS_PORT=3000
 # REQUIRED: manually defined by user, or default will be used
 BLOCKER_LOG_LEVEL={{ webportal_common_config.blocker_log_level }}
 
-# OPTIONAL: manually defined by user, can be blank
-BLOCKER_NGINX_CACHE_PURGE_LIST={{ webportal_common_config.blocker_nginx_cache_purge_list | default('') }}
-BLOCKER_NGINX_CACHE_PURGE_LOCK={{ webportal_common_config.blocker_nginx_cache_purge_lock | default('') }}
+# OPTIONAL: manually defined by user
+BLOCKER_PORTALS_SYNC={{ webportal_common_config.blocker_portals_sync }}
```
MSevey, line 231: If is it optional we should add a default value here so that Ansible doesn't complain when it isn't defined. 

Is this a boolean? String?

You can copy the syntax used in the lines you removed to add a default value.

MSevey, line 231: If is it optional we should add a default value here so that Ansible doesn't complain when it isn't defined.

Is this a boolean? String?

You can copy the syntax used in the lines you removed to add a default value.

MSevey (changes requested):

playbooks/templates/.env.j2
firyx: Are `BLOCKER_NGINX_...` used somewhere?
Can we safely remove them?
We need to keep backwards compatibility of playbooks if someone (we or portal operators) deploys older/stable versions of the stack.

playbooks/templates/.env.j2
```diff
@@ -227,9 +227,17 @@ SKYNET_ACCOUNTS_PORT=3000
...
 BLOCKER_LOG_LEVEL={{ webportal_common_config.blocker_log_level }}
 
-# OPTIONAL: manually defined by user, can be blank
-BLOCKER_NGINX_CACHE_PURGE_LIST={{ webportal_common_config.blocker_nginx_cache_purge_list | default('') }}
-BLOCKER_NGINX_CACHE_PURGE_LOCK={{ webportal_common_config.blocker_nginx_cache_purge_lock | default('') }}
+# List of portal urls from which to sync the blocklist periodically
+#
+# OPTIONAL: manually defined by user
+BLOCKER_PORTALS_SYNC={{ webportal_common_config.blocker_portals_sync |
+default('')}}
```
firyx, line 234:
```suggestion
default('') }}
```

firyx (changes requested):

playbooks/templates/.env.j2
```diff
@@ -227,9 +227,17 @@ SKYNET_ACCOUNTS_PORT=3000
...
+#
+# OPTIONAL: manually defined by user
+BLOCKER_PORTALS_SYNC={{ webportal_common_config.blocker_portals_sync |
+default('') }}
+
+# OPTIONAL: manually defined by user
+BLOCKER_NGINX_CACHE_PURGE_LOCK={{
+webportal_common_config.blocker_nginx_cache_purge_lock | default('') }}
+BLOCKER_NGINX_CACHE_PURGE_LIST={{
+webportal_common_config.blocker_nginx_cache_purge_list | default('') }}
```
kwypchlo, line 240: does this work with that new line? is there a reason why you changed those settings order (looking at changes)?

playbooks/templates/.env.j2
```diff
@@ -227,9 +227,17 @@ SKYNET_ACCOUNTS_PORT=3000
...
-# OPTIONAL: manually defined by user, can be blank
-BLOCKER_NGINX_CACHE_PURGE_LIST={{ webportal_common_config.blocker_nginx_cache_purge_list | default('') }}
-BLOCKER_NGINX_CACHE_PURGE_LOCK={{ webportal_common_config.blocker_nginx_cache_purge_lock | default('') }}
+# List of portal urls from which to sync the blocklist periodically
+#
+# OPTIONAL: manually defined by user
+BLOCKER_PORTALS_SYNC={{ webportal_common_config.blocker_portals_sync |
+default('') }}
+
+# OPTIONAL: manually defined by user
```
firyx, line 236: Maybe we should comment those as:
Kept for Ansible playbooks backwards compatibility with older Blocker versions.

firyx (changes requested):
Also changelog item is missing.

kwypchlo (approved):

firyx (approved):

PR #264, opened 2022-03-22
move portal modules outside of server config

MSevey:
# PULL REQUEST

## Overview
Update how backwards compatibility is handled with portal_modules.

### DIFF
--- a/playbooks/group_vars/webportals.yml
+++ b/playbooks/group_vars/webportals.yml
@@ -178,7 +178,7 @@ webportal_allowance:
 # TODO: move to role default vars
 webportal_server_config_defaults:
   domain_name: "{{ ansible_host }}"
-  portal_modules: "{{ portal_modules | default('') }}"
+  portal_modules: ""
   sia_wallet_password: ""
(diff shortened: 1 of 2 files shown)


playbooks/group_vars/webportals.yml
```diff
@@ -178,7 +178,7 @@ webportal_allowance:
 # TODO: move to role default vars
 webportal_server_config_defaults:
   domain_name: "{{ ansible_host }}"
-  portal_modules: "{{ portal_modules | default('') }}"
+  portal_modules: ""
```
firyx, line 181: Shouldn't we leave this out?

If we already have it in LastPass, this default will not be used.

If we do not have it, operator should define it in `hosts.ini` and we shouldn't pollute LastPass with default empty value anymore.

firyx (changes requested):

firyx (approved):

ro-tex (approved):

ro-tex (approved):

firyx (approved):

PR #265, opened 2022-03-22
do not prune docker data in portal-role-task

kwypchlo:
Running setup-portal-following will execute "prunning old docker data" which means that if we run a deploy on a machine, it will remove all docker data older than 4 hours from that machine and following deploy playbook will not take advantage of any cached builds. This is why we removed pruning the docker data step directly before the build step.

### DIFF
--- a/playbooks/tasks/portal-role-task.yml
+++ b/playbooks/tasks/portal-role-task.yml
@@ -111,9 +111,6 @@
   # Don't rebuild when initializing sia with existing seed is in progress
   when: create_env_file_result.changed or sia_container_info_result.container == None or not sia_container_info_result.container.State.Running
 
-- name: Include prunning old docker data
-  include_tasks: tasks/portal-docker-prune-old-data.yml
-
 - name: Check wallet
   command: docker exec sia siac wallet
   register: sia_wallet_result


playbooks/tasks/portal-role-task.yml
```diff
@@ -111,9 +111,6 @@
   # Don't rebuild when initializing sia with existing seed is in progress
   when: create_env_file_result.changed or sia_container_info_result.container == None or not sia_container_info_result.container.State.Running
 
-- name: Include prunning old docker data
-  include_tasks: tasks/portal-docker-prune-old-data.yml
-
```
MSevey, line 116: Instead of deleting this, I would just add a when condition

```yml
when: not deploy_after_setup
```

Since when we are setting up the portal, we might run into issues or debugging something where we want to be rebuilding the docker images and not using potentially bad docker caches.

MSevey (changes requested):

playbooks/tasks/portal-role-task.yml
kwypchlo: Docker is pretty good with caching, the only reason why we added pruning was because we were running out of disk space after rebuilding all images during deploy phase tens of times. I think it's safe to assume that if someone is running setup-following he will follow with deploy at some point that will handle any pruning old data after the relevant pieces are built. Also I don't think the disk space reasoning applies in this particular place because all that cache that we're pruning, we'll have to recreate again anyway in the deploy task - it's just a waste of processing time (and in any case the disk space used for cache here is tiny and should not have any impact on overall disk space usage).

MSevey: I wasn't suggesting keeping this here due to disk usage concerns during portal-setup-following.

I was concerned with future runs of portal-setup-following needing to build fresh docker images. 

But if that isn't a concern then this is fine.

MSevey (approved):

firyx (approved):

PR #268, opened 2022-03-23
Yamllint and syntax check

MSevey:
# PULL REQUEST

## Overview
Add Yamllint and syntax checking to repo with Makefile and github action.

A follow up will be to enforce the lint github action as a requirement for PRs. 

## Checklist
 - [x] Changelog file created

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,9 +1,9 @@
-name: yamllint test
+name: Lint
 
 on: push
 
 jobs:
-  test:
+  lint:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
@@ -17,4 +17,14 @@ jobs:
         run: pip install yamllint
 
       - name: Lint YAML files
-        run: yamllint --no-warnings .
+        run: make lint
+
+  syntax_check:
+    runs-on: ubuntu-latest
+    steps:
+      - uses:
+          - actions/checkout@v2
+          - docker/build-push-action@v1
+
+      - name: Check playbooks with --syntax-check
+        run: make syntax-check
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -24,10 +24,5 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
-      - name: Set up Python
-        uses: actions/setup-python@v2
-        with:
-          python-version: 3.8
-
       - name: Check playbooks with --syntax-check
         run: make syntax-check
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -22,9 +22,12 @@ jobs:
   syntax_check:
     runs-on: ubuntu-latest
     steps:
-      - uses:
-          - actions/checkout@v2
-          - docker/build-push-action@v1
+      - uses: actions/checkout@v2
+
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.8
 
       - name: Check playbooks with --syntax-check
         run: make syntax-check
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,16 @@
-.PHONY: lint fmt
+.PHONY: dependencies clean lint fmt syntax-check
 
 all: dependencies
+
+# clean removes and re-initializes the directories that have their contents
+# ignored by git. This will force a re-installation of the requirements, roles,
+# etc.
+clean:
+	rm -rf my-logs
+	mkdir my-logs
+	touch my-logs/.empty-file
+	rm -rf ansible_collections
+	rm -rf roles
 
 dependencies:
 	pip install yamllint
@@ -18,3 +28,13 @@ dependencies:
 # https://yamllint.readthedocs.io/
 lint:
 	yamllint --no-warnings .
+
+# syntax-check statically checks the syntax for all the playbooks in playbooks
+# directory
+syntax-check:
+	for f in $$(ls ./playbooks); do \
+		if ! [ -d ./playbooks/$$f ]; \
+		then \
+			./scripts/syntax-check.sh ./playbooks/$$f || exit 1; \
+		fi \
+	done
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,20 @@
+.PHONY: lint fmt
+
+all: dependencies
+
+dependencies:
+	pip install yamllint
+
+# yamlfix is a helper tool that will fix common yaml errors. If we can enable it
+# that would help ensure code standards but we need tests first to ensure the
+# updates don't break things. Until then we need to address the lint errors
+# manually.
+#
+# https://lyz-code.github.io/yamlfix/
+#	pip install yamlfix
+#fmt:
+#	yamlfix .
+
+# https://yamllint.readthedocs.io/
+lint:
+	yamllint --no-warnings .
--- a/playbooks/host_vars/crap.yml
+++ b/playbooks/host_vars/crap.yml
@@ -18,4 +18,5 @@ security_fail2ban_enabled: True
 # Custom ufw rules
 host_ufw_rules:
   # Vagrant SSH port
-  - { rule: 'limit', port: '2222', proto: 'tcp', direction: 'in' }
+  # yamllint disable-line rule:braces
+  - { rule: "limit", port: "2222", proto: "tcp", direction: "in" }
--- a/playbooks/tasks/portal-role-task-mongo-check-docker-service.yml
+++ b/playbooks/tasks/portal-role-task-mongo-check-docker-service.yml
@@ -40,7 +40,7 @@
   when: not mongo_service_ok
 
 - name: Wait for mongo container initialization before checking status
-  ansible.builtin.shell: 'docker logs mongo | grep ''"Waiting for connections"'''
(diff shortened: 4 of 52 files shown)


kwypchlo (approved):

playbooks/tasks/portal-role-task-mongo-check-docker-service.yml
```diff
@@ -41,7 +40,7 @@
   when: not mongo_service_ok
 
 - name: Wait for mongo container initialization before checking status
-  ansible.builtin.shell: "docker logs mongo | grep '\"Waiting for connections\"'"
+  ansible.builtin.shell: 'docker logs mongo | grep ''"Waiting for connections"'''
```
MSevey, line 43: need to undo this I think

playbooks/tasks/portal-role-task-mongo-check-docker-service.yml
```diff
@@ -81,7 +80,7 @@
   vars:
     mongodb_shell:
       eval: "db.serverStatus()"
-      until: "'\"ok\" : 1' in mongo_shell_result.transformed_output"
+      until: '''"ok" : 1'' in mongo_shell_result.transformed_output'
```
MSevey, line 83: same

playbooks/tasks/portal-role-task-mongo-setup-replicaset-init.yml
```diff
@@ -42,6 +41,6 @@
   fail:
     msg: |
       MongoDB replicaset initialization was not successful:
-      
+
       Mongo shell result: {{ mongo_shell_result }}
-  when: "mongo_shell_result.transformed_output | default('none output') != '{ \"ok\" : 1 }'"
+  when: 'mongo_shell_result.transformed_output | default(''none output'') != ''{ "ok" : 1 }'''
```
MSevey, line 46: same

playbooks/tasks/portal-role-task-mongo-setup.yml
```diff
@@ -90,7 +89,7 @@
   when: not mongo_service_ok
 
 - name: Wait for mongo container initialization before checking status
-  ansible.builtin.shell: "docker logs mongo | grep '\"Waiting for connections\"'"
+  ansible.builtin.shell: 'docker logs mongo | grep ''"Waiting for connections"'''
```
MSevey, line 92: same

playbooks/tasks/portal-role-task-mongo-setup.yml
```diff
@@ -136,7 +135,7 @@
   vars:
     mongodb_shell:
       eval: "db.serverStatus()"
-      until: "'\"ok\" : 1' in mongo_shell_result.transformed_output"
+      until: '''"ok" : 1'' in mongo_shell_result.transformed_output'
```
MSevey, line 138: same

playbooks/tasks/portal-role-task-mongo-setup-replicaset-init.yml
firyx, line 46: Is this going to work correctly?
I think that single quotes will make it a string (not an expression) and will always evaluate to the same bool value.
I'm not even sure now that double quotes worked correctly.
Can we test it with a simple task?

Makefile
```diff
@@ -0,0 +1,40 @@
+.PHONY: dependencies clean lint fmt syntax-check
+
+all: dependencies
+
+# clean removes and re-initializes the directories that have their contents
+# ignored by git. This will force a re-installation of the requirements, roles,
+# etc.
+clean:
+	rm -rf my-logs
```
firyx, line 9: If I want to run linting locally, I don't want to delete all logs everytime. To force re-installation of the requirements & co. maybe just deleting `my-logs/requirements-installed.txt` file should be enough.

scripts/syntax-check.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+
+# This script is a helper for checking the syntax of any ansible playbook
+
+# Set command and arguments
+cmd=ansible-playbook
+args="--syntax-check $@"
```
firyx, line 7: IIRC syntax check doesn't go down to included files/tasks, I think it will stay just at first playbook level.

We could maybe do a syntax check for all playbook files and then for all task (and is possible also for template/group var/host var) files iteratively to have 100% coverage.

firyx (changes requested):

Makefile
MSevey: I would recommend a separate command to just remove the requirements file. 
`clean` should clean everything from the repo, as if it was recently cloned. 

either way, recommend this as a follow up if we want to discuss changing this more.

scripts/syntax-check.sh
MSevey: you are right, unfortunately you can't call syntax check on the tasks files because they are not full playbooks so they fail. so we can only use it on playbooks. 

I'm hoping that ansible-lint will be able to handle those files.

playbooks/tasks/portal-role-task-mongo-setup-replicaset-init.yml
MSevey: sneaky sneaky, I thought I reverted this but my code editor must have reset it on save

Makefile
```diff
@@ -0,0 +1,40 @@
...
+
+# syntax-check statically checks the syntax for all the playbooks in playbooks
+# directory
+syntax-check:
+	for f in $$(ls ./playbooks); do \
+		if ! [ -d ./playbooks/$$f ]; \
+		then \
+			./scripts/syntax-check.sh ./playbooks/$$f || exit 1; \
+		fi \
+	done
```
kwypchlo, line 40: is 8 spaces indent some sort of a standard in Makefiles? looks really spread out, I usually use either 2 for any code, yaml, json etc - your yml files also have 2 spaces indents

playbooks/host_vars/crap.yml
```diff
@@ -18,4 +18,5 @@ security_fail2ban_enabled: True
 # Custom ufw rules
 host_ufw_rules:
   # Vagrant SSH port
-  - { rule: 'limit', port: '2222', proto: 'tcp', direction: 'in' }
+  # yamllint disable-line rule:braces
```
kwypchlo, line 21: what is yamllint suggesting in this case since we're disabling the rule ?

playbooks/host_vars/crap.yml
```diff
@@ -18,4 +18,5 @@ security_fail2ban_enabled: True
...
+  - { rule: "limit", port: "2222", proto: "tcp", direction: "in" }
```
kwypchlo, line 22: what is yamllint suggesting in this case since we're disabling the rule ?

scripts/lib/ansible-executor.sh
```diff
@@ -109,4 +110,12 @@ echo "Executing:"
 echo "    $cmd $args"
 echo "in a docker container $ansiblecm_container..."
 
-docker exec -it $ansiblecm_container $cmd $args
+# Normally we want to use the -it flag, but with GitHub actions you will get an
+# error for 'not a tty device'. This is simply saying that the color formatting
+# and other interactive features of the terminal are not supported, so we do not
+# submit those flags if that is the case. 
+it_flags="-it"
+if ! [ -z ${github_action+x} ]; then it_flags=""; fi
```
kwypchlo, line 118: `-it` means "interactive" so it also breaks if there is any interaction required from user, ie prompt

This will break those 2 flows you mentioned recently that have prompts if ran from non-tty device.

.github/workflows/lint.yml
```diff
@@ -0,0 +1,28 @@
...
+      - name: Lint YAML files
+        run: make lint
+
+  syntax_check:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Check playbooks with --syntax-check
+        run: make syntax-check
```
kwypchlo, line 28: is there a reason why are you using make instead of github pipeline directly?

MSevey: I'm using make so that we ensure consistency of running the command. I don't want people running the command one way locally and then a different way on the CI.

scripts/lib/ansible-executor.sh
MSevey: It doesn't break because `--syntax-check` unfortunately doesn't check included tasks, which are the tasks you are referring to.

Makefile
MSevey: yea that is what tab does for the makefile. not sure why but our other Makefiles have the same spacing.

https://gitlab.com/SkynetLabs/skyd/-/blob/master/Makefile#L222

playbooks/host_vars/crap.yml
MSevey: it wants `{key: value}` instead of `{ key: value }`

personally, I think the extra space makes it more readable, plus I can't figure out what setting in my code editor is forcing the spaces because if I delete them they are auto-added back in haha

kwypchlo: well I think you probably want to fix that locally for you instead then :)

MSevey: I created a follow up issue. this shouldn't block this PR as it is a stylistic preference.

kwypchlo (approved):

firyx (approved):

PR #277, opened 2022-03-29
Fix takedowns

firyx:
# PULL REQUEST

## Overview

This PR fixes 2 takedown issues:
- `portal-setup-following` with enabled deploy dodn't honor portal takedown
- portal takedown run only `sia` docker service, not `mongo`

Moreover this PR enables portal deploys and restarts to takedown portals.

## Checklist
 - [x] Changelog file created


playbooks/portals-takedown.yml
```diff
@@ -42,15 +42,17 @@
       include_tasks: tasks/portal-docker-compose-files-get-wanted.yml
 
     # Start just sia docker service
-    - name: Start just sia docker service
+    - name: Start just mongo and sia docker services
       community.docker.docker_compose:
         project_src: "{{ webportal_dir }}"
         files: "{{ webportal_docker_compose_files_wanted }}"
         build: False
```
kwypchlo, line 49: does it mean we're not rebuilding ? I think this should be set to true ?

MSevey (dismissed):

playbooks/portals-takedown.yml
firyx: On takedown playbook we are not rebuilding, we just start existing mongo and sia services. Similar on restart playbooks, we are not rebuilding on them. We are rebuilding on deploys.

kwypchlo: ok but we still can run setup-portal following and it will update the stack but start only sia and mongo right?

firyx: Yes, if the portal is in takedown group.

peterjan (approved):
LGTM 👍

PR #284, opened 2022-04-05
Mongo setup replicaset 2

firyx:
# PULL REQUEST

## Checklist
 - [ ] Changelog file created


playbooks/tasks/portal-mongo-replicaset-check-majority.yml
```diff
@@ -46,7 +46,9 @@
     # replicaset majority check, it doesn't matter that we stop the rest of
     # services.
     - block:
-        - name: Include checking mongo replicaset majority
+        # Get MongoDB replicaset status details
+
```
MSevey, line 50: nit: move newline to above comment.

playbooks/tasks/portal-role-task-mongo-replicaset-ensure-voting-non-voting-member.yml
```diff
@@ -0,0 +1,107 @@
...
+
+- name: Include getting MongoDB voting members count
+  include_tasks: tasks/portal-role-task-mongo-shell-eval-on-delegated-host.yml
+  vars:
+    mongodb_shell_delegated:
+      host: "{{ mongo_primary_host }}"
+      eval: |
+        print(rs.config().members.filter(m => m.votes > 0).length)
+
+- name: Set MongoDB replicaset voting members
```
MSevey, line 73: Kind of a `nit` but also without `count` I would read this task as the entire act of setting the voting members. 

```suggestion
- name: Set MongoDB replicaset voting members count
```

playbooks/tasks/portal-role-task-mongo-replicaset-ensure-voting-non-voting-member.yml
```diff
@@ -0,0 +1,107 @@
...
+  set_fact:
+    voting_members_count: "{{ mongo_shell_delegated_result.transformed_output }}"
+
+- name: Set MongoDB votes and priority
```
MSevey, line 77:
```suggestion
# Determine if the new mongo node should be a voting or 
# non-voting member of the replicaset by checking if there are 
# already `mongo_max_voting_members`
- name: Determine if MongoDB node should be voting or non-voting
```

playbooks/tasks/portal-role-task-mongo-replicaset-get-primary.yml
```diff
@@ -3,25 +3,28 @@
...
-    loop_var: "host_item"
+- name: Get MongoDB replicaset primary member
+  include_tasks: tasks/portal-role-task-mongo-shell-eval.yml
+  vars:
+    mongodb_shell:
+      # Due to yaml escaping following eval must use ':', not ":".
+      # isMaster.primary returns string in format '<FQDN>:<port>', we need to
+      # extract only the first subdomain which is Ansible host.
+      eval: |
+        rs.isMaster().primary.split('.')[0]
```
MSevey, line 21: based on the comment shouldn't it be this:
```suggestion
        rs.isMaster().primary.split(':')[0]
```

playbooks/tasks/portal-role-task.yml
```diff
@@ -138,7 +134,14 @@
 
 - name: Set Sia wallet seed defined
   set_fact:
-    sia_wallet_seed_defined: "{{ webportal_server_config.sia_wallet_seed is defined and webportal_server_config.sia_wallet_seed != None }}"
+    sia_wallet_seed_defined: >-
+      {{
```
MSevey, line 138:
```suggestion
      "{{
```

MSevey (changes requested):

playbooks/tasks/portal-mongo-replicaset-check-majority.yml
firyx: There are 2 conventions that we follow:

1) after `- block:` we always start code or comments without newlines inbetween.

2) Comment + optional newline + task code
  - If the comment is related only the one following task, there is no newline between the comment and the task (comment is sticked to the task).
  - If the comment is related to several tasks that follow, there is a newline between the comment and the tasks that follow.

Example:

```
- block:
  # Get abc
  - name: Get abc
    set_fact:
      abc: "{{ some code to get abc is here }}"

- block:
  # Get something

  - name: Run something preparation command
    command: ...
    register: something_preparation_command_result

  - name: Get something out of the result
    set_fact:
      something: "{{ some code to get something out of something_preparation_command_result }}"
```

playbooks/tasks/portal-role-task-mongo-replicaset-get-primary.yml
firyx: That suggestion is not correct.

`rs.isMaster().primary` returns e.g. `us-ny-1.siasky.net:27017`.
We need to extract `Ansible host` which is just `us-ny-1` (not FQDN `us-ny-1.siasky.net`).

playbooks/tasks/portal-role-task.yml
firyx: Good catch.
Actually the quotation marks should be removed from the end of the block, not from the beginning.

playbooks/tasks/portal-role-task-mongo-replicaset-get-primary.yml
firyx: I have updated the comment not to mention `FQDN` but subdomain.

playbooks/tasks/portal-role-task-mongo-replicaset-ensure-voting-non-voting-member.yml
```diff
@@ -0,0 +1,110 @@
...
+  delegate_facts: True
+
+- name: Include checking if MongoDB node is part of the replicaset
+  include_tasks: tasks/portal-role-task-mongo-shell-eval-on-delegated-host.yml
+  vars:
+    mongodb_shell_delegated:
+      host: "{{ mongo_primary_host }}"
+      eval: |
+        rs.config().members.forEach(m => {
+          if (m.host == '{{ ansible_host }}:{{ subcrap_mongo_port | default(default_mongo_port | string) }}')
```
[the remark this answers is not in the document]
MSevey: yea agreed, is here not another way we can define a custom mongo port such that it is
```yml
{{ custom_mongo_port | default(default_mongo_port) | string}}
```

playbooks/tasks/portal-role-task-mongo-replicaset-get-primary.yml
MSevey: I guess I'm not following the comment. What is it referring to? Where do we need to be referring to `':'`? 

Right now the comment s more confusing the helpful IMO.

firyx: oh, I have removed the complete comment `# Due to yaml escaping following eval must use ':', not ":".`
because the eval scripts has changed since the comment was added, the eval script doesn't contain `:` character anymore.

playbooks/tasks/portal-role-task-mongo-replicaset-ensure-voting-non-voting-member.yml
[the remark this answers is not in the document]
firyx: renamed to `custom_mongo_port`

playbooks/tasks/portal-role-task-mongo-shell-eval.yml
```diff
@@ -6,9 +6,8 @@
 - name: Run mongo shell command
   community.mongodb.mongodb_shell:
     # Use secure localhost connection unless on subcrap test cluster
-    login_host: "{{ 'localhost' if (subcrap_mongo_port is not defined) else ansible_host }}"
-    # Use default port unless on subcrap test cluster
-    login_port: "{{ subcrap_mongo_port | default(default_mongo_port | string) }}"
+    login_host: "{{ 'localhost' if (custom_mongo_port is not defined) else ansible_host }}"
```
kwypchlo, line 9: I do not understand the logic here, why does `custom_mongo_port` affect the host ?

MSevey (approved):

playbooks/tasks/portal-role-task-mongo-shell-eval.yml
kwypchlo: I still don't get it I think. Why would we use localhost instead of ansible_host, it doesn't seem like a valid setup case.

firyx: If we don't need to test mongo on virtual machines there would be only:
`login_host: "localhost"`

The `localhost` in this task means a mongo container under which we run the mongo shell script (`mongo_cmd: "docker exec mongo mongo"`).

With `localhost` we are sure that mongo setup scripts stay and don't in any case leave/leak out of mongo container.

When there is a need to test mongo cluster which is setup e.g. on `crap` host machine as `subcrap` virtual machines we can loose up the security a bit and target `subcrap` mongo nodes we are using `ansible_host` (instead of `localhost` which targets `crap` mongo node).

OTOH there was a coupling between using custom mongo port and using non-localhost target, i.e. using custom mongo port assumed subcrap testing of mongo cluster. I decoupled it so that custom mongo port is now independent of mongo cluster tests via usage of `use_custom_mongo_host` variable.

kwypchlo (approved):

MSevey (approved):

PR #285, opened 2022-04-05
Reporting Env Variables

peterjan:
# PULL REQUEST

## Overview
This PR adds the necessary environment variables for the reporting feature.

## Example for Visual Changes
N/A

## Checklist
 - [x] Changelog file created

## Issues Closed
N/A

### DIFF
--- a/playbooks/group_vars/webportals.yml
+++ b/playbooks/group_vars/webportals.yml
@@ -195,6 +195,7 @@ webportal_server_config_defaults:
 webportal_common_config_defaults:
   s3_backup_path: ""
   abuse_log_level: "info"
+  abuse_ncmec_reporting_enabled: False
   accounts_log_level: "info"
   blocker_log_level: "info"
   cookie_hash_key: "{{ lookup('pipe','openssl rand -hex 32') }}"
@@ -204,6 +205,7 @@ webportal_common_config_defaults:
   skynet_db_port: "{{ default_mongo_port }}"
   skynet_db_replicaset: "{{ portal_cluster_id }}"
   skynet_db_user: "admin"
+  ncmec_debug: True
 
 # TODO: move to role default vars
 # TODO: default to: False
--- a/playbooks/group_vars/webportals.yml
+++ b/playbooks/group_vars/webportals.yml
@@ -205,7 +205,6 @@ webportal_common_config_defaults:
   skynet_db_port: "{{ default_mongo_port }}"
   skynet_db_replicaset: "{{ portal_cluster_id }}"
   skynet_db_user: "admin"
-  ncmec_debug: True
 
 # TODO: move to role default vars
 # TODO: default to: False
--- a/playbooks/templates/.env.j2
+++ b/playbooks/templates/.env.j2
@@ -280,6 +280,10 @@ NCMEC_USERNAME={{ webportal_common_config.ncmec_username }}
 NCMEC_PASSWORD={{ webportal_common_config.ncmec_password }}
 NCMEC_DEBUG={{ webportal_common_config.ncmec_debug }}
 
+NCMEC_REPORTER_FIRSTNAME={{ webportal_common_config.ncmec_reporter_firstname }}
+NCMEC_REPORTER_LASTNAME={{ webportal_common_config.ncmec_reporter_lastname }}
+NCMEC_REPORTER_EMAIL={{ webportal_common_config.ncmec_reporter_email }}
+
 {% endif %}
 {% endif %}
 # TODO: Add serverlist script vars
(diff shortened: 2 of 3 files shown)


playbooks/templates/.env.j2
```diff
@@ -268,6 +268,17 @@ ABUSE_LOG_LEVEL={{ webportal_common_config.abuse_log_level }}
 # REQUIRED: Hardcoded, not configurable
 ABUSE_SPONSOR="Abuse Scanner"
 
+# Reporting related environment variables
+#
+# OPTIONAL: manually defined by user, defaults to false
+ABUSE_NCMEC_REPORTING_ENABLED={{
+webportal_common_config.abuse_ncmec_reporting_enabled | default('false') }}
+#
+# REQUIRED: manually defined by user, only if reporting is enabled
```
MSevey, line 277: REQ: A user could have the abuse scanner enabled but not the NCMEC reporting correct?

If so then these should be in an if condition based on `webportal_common_config.abuse_ncmec_reporting_enabled` otherwise ansible is going to complain and error when these aren't defined. 
```
{% if webportal_common_config.abuse_ncmec_reporting_enabled %}
...
{% endif %}

**Review by MSevey — DISMISSED** [review]

File: playbooks/templates/.env.j2 [hunk 2]
```diff
@@ -268,6 +268,19 @@ ABUSE_LOG_LEVEL={{ webportal_common_config.abuse_log_level }}
...
+{% if webportal_common_config.abuse_ncmec_reporting_enabled %}
```
firyx, line 278:
```suggestion
{% if webportal_common_config.abuse_ncmec_reporting_enabled | default(False) %}
```

playbooks/templates/.env.j2
peterjan: right, thanks!

playbooks/templates/.env.j2
```diff
@@ -268,6 +268,19 @@ ABUSE_LOG_LEVEL={{ webportal_common_config.abuse_log_level }}
...
+NCMEC_USERNAME={{ webportal_common_config.ncmec_username }}
+NCMEC_PASSWORD={{ webportal_common_config.ncmec_password }}
+NCMEC_DEBUG={{ webportal_common_config.ncmec_debug | default('true') }}
+{% endif %}
```
firyx, line 282: See other `{% if %}` blocks how we handle whitespace (we don't want to add double empty new lines), we enclose trainilng new line inside the block before `{% endif %}`.

firyx (changes requested):

playbooks/templates/.env.j2
```diff
@@ -268,6 +268,20 @@ ABUSE_LOG_LEVEL={{ webportal_common_config.abuse_log_level }}
...
+webportal_common_config.abuse_ncmec_reporting_enabled }}
+#
+# REQUIRED: manually defined by user, only if reporting is enabled
+{% if webportal_common_config.abuse_ncmec_reporting_enabled %}
+NCMEC_USERNAME={{ webportal_common_config.ncmec_username }}
+NCMEC_PASSWORD={{ webportal_common_config.ncmec_password }}
+NCMEC_DEBUG={{ webportal_common_config.ncmec_debug }}
+
+{% endif %}
+
```
firyx, line 284: This line should be removed.

firyx (changes requested):

firyx (approved):

firyx (approved):

playbooks/group_vars/webportals.yml
```diff
@@ -204,6 +205,7 @@ webportal_common_config_defaults:
   skynet_db_port: "{{ default_mongo_port }}"
   skynet_db_replicaset: "{{ portal_cluster_id }}"
   skynet_db_user: "admin"
+  ncmec_debug: True
```
ChrisSchinnerl, line 208: Do we need to set this when the reporting is disabled anyway?

peterjan: The way I understand it is that we want to list the defaults here, in this file because it will autosave the value in LastPass and it's easier to find all config values. If reporting is disabled this variable simply won't be used and it won't show up in the env file.

But yeah other than that I also don't really know how to respond to that question. It might be redundant and personally, I also don't really see why default variables should be listed here rather than in the template itself.

peterjan: Talked it over with Fil. Probably the best approach is to leave this one out of the default vars indeed.
Updated it.

firyx (approved):

firyx (approved):

PR #287, opened 2022-04-06
Autogenerate mongo db mgkey

firyx:
# PULL REQUEST

## Overview

This PR is a preparation for complete MongoDB setup. It autogenerates `mongo_db_mgkey` and saves it to LastPass if the variable is not defined or if the portal operator already defined it in LastPass, but kept it empty.

This PR also improves checking whether MongoDB replicaset is already initialized.

## Checklist
 - [x] Changelog file created


playbooks/tasks/portal-configs-load.yml
```diff
@@ -2,6 +2,7 @@
 # Get portal configs
 # - Load portal configs from secrets backend (e.g. LastPass)
 # - Set default values (for unset keys)
+# - Set mongodb mgkey (if defined but empty)
```
ro-tex, line 5:
```suggestion
# - Set mongodb mgkey (if either undefined or defined but empty)
```

ro-tex (dismissed):

playbooks/tasks/lastpass-save-cluster-config.yml
```diff
@@ -49,10 +49,44 @@
...
+          set_fact:
+            # Add prefix to each mgkey line
+            indented_mgkey: >-
+              {{ [prefix]
+              | product(webportal_common_config_last.mongo_db_mgkey.splitlines())
+              | map('join')
+              | list
+              | join('\n') }}
+
+        - name: Update last config with a dummy mongo mgkey value
```
MSevey, line 67: Feels like this needs a comment explaining why we need a dummy value.

MSevey, line 67: REQ:  This needs a comment. Why are we adding a dummy value to the mgkey value?

playbooks/tasks/lastpass-save-cluster-config.yml
```diff
@@ -49,10 +49,44 @@
...
+              {{ webportal_common_config_last
+              | combine({'mongo_db_mgkey': 'dummy'}) }}
+
+        - name: Create last config yaml string
+          set_fact:
+            common_config_last_yaml: >-
+              {{
+                webportal_common_config_last
+                | to_nice_yaml(width=2048)
+                | replace(
```
MSevey, line 79: REQ: code like this needs comments. Is using replace like this some sort of work around? It doesn't make sense why we would write a dummy value only to overwrite it immediately after.

playbooks/tasks/portal-role-task-mongo-setup-docker-service.yml
```diff
@@ -6,9 +6,11 @@
   assert:
     that:
       - >-
-        {{ webportal_common_config.mongo_db_mgkey is defined and
-        webportal_common_config.mongo_db_mgkey is not none and
-        webportal_common_config.mongo_db_mgkey != '' }}
+        {{
+          webportal_common_config.mongo_db_mgkey is defined
```
MSevey, line 10: nit: If they are all `and` conditionals, i think you can just do a list right?
```yaml
- condition 1
- condition 2
- condition 3
```
I believe ansible interperats lists as `and` conditions by default.

MSevey (changes requested):

playbooks/tasks/portal-configs-load.yml
firyx: That is not exact:

It is like this:

`# - Set default values (for unset keys)`
- key is not set (= undefined) => the value is set here

`# - Set mongodb mgkey (if defined but empty)`
- key is set (= defined) but value is not set => the value is set here

playbooks/tasks/lastpass-save-cluster-config.yml
firyx: Added an explanation comment before the whole block incl. examples.

playbooks/tasks/lastpass-save-cluster-config.yml
firyx: See above.

playbooks/tasks/portal-role-task-mongo-setup-docker-service.yml
firyx: Updated.

playbooks/tasks/portal-configs-load.yml
ro-tex: OK.

MSevey (approved):

PR #295, opened 2022-04-22
Add loading of custom config files in ansible-private

### LINKED ISSUE #293: Didn't we agreed that we will not introduce any more group vars files and will use hosts.ini instead?
Didn't we agreed that we will not introduce any more group vars files and will use hosts.ini instead?

_Originally posted by @firyx in https://github.com/SkynetLabs/ansible-playbooks/pull/292#discussion_r856228466_

From discord
```
we have some options:
if we want to keep custom configs out of playbooks repo
and they look ugly (not nice to read) in hosts.ini
we could define in hosts.ini (yaml config) filename(s) to load
and define these custom configs in ansible-private/custom-configs/....yml as a yaml file(s)
so we get good things from both worlds (ini and yaml)
in playbooks we just load vars from yaml files defined in hosts.ini,
I think, the playbook can use vars_files 
```

MSevey:
# PULL REQUEST

## Overview
Add the ability to load custom variable config files. These can be defined in hosts.ini

Included tasks for setup following and deploy.

These are the updates to ansible-private to support these changes
<https://github.com/SkynetLabs/ansible-private/pull/32>

## Checklist
 - [x] Changelog file created

## Issues Closed

Closes #293


ro-tex (dismissed):

playbooks/portals-setup-following.yml
```diff
@@ -25,6 +25,9 @@
     - name: Include checking server OS compatibility
       include_tasks: tasks/check-server-os-compatibility.yml
 
+    - name: Load Custom Variable File
```
firyx, line 28: We need to load vars also e.g. in `portals-setup-initial` and basically in every playbook and play where we set `lastpass_required: True`.

We should be able to do it simply using `vars_file`:
https://docs.ansible.com/ansible/latest/user_guide/playbooks_variables.html#defining-variables-in-included-files-and-roles

```
  vars:
    lastpass_required: True
  vars_file:
    - "{{ common_vars_file | default(omit) }}"
    - "{{ custom_vars_file | default(omit) }}"
```

Note that e.g. `portals-setup-following` playbook consists from 4 plays, we need to set it on all 4 plays.
And while `default(omit)` should work we should test it works as expected with `vars_file`.

firyx (changes requested):

firyx (approved):

peterjan (approved):

PR #299, opened 2022-04-28
pin pip package versions

kwypchlo:
- pin specific versions of packages instead of using latest
- also according to [ansible pip docs](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/pip_module.html) we don't need to define loop, we can just define a number of packages we want installed

### DIFF
--- a/playbooks/deploy-machines-setup.yml
+++ b/playbooks/deploy-machines-setup.yml
@@ -52,10 +52,8 @@
 
     - name: Install Docker SDK for Python
       pip:
-        name: "{{ item }}"
-        state: present
-      loop:
-        - docker
+        name:
+          - docker==5.0.3
 
     - name: Set timezone
       community.general.timezone:
--- a/playbooks/tasks/portal-role-task.yml
+++ b/playbooks/tasks/portal-role-task.yml
@@ -307,13 +307,11 @@
 - block:
     - name: Install Python modules for health checks
       pip:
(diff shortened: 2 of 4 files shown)


kwypchlo:
> > also according to [ansible pip docs](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/pip_module.html) we don't need to define loop, we can just define a number of packages we want installed
> 
> the loop was to refactor the packages since they were all using latest. so yes, if you are defining different images for all the packages it doesn't make sense to use the loop.

Can you clarify? We could have used "name" list even without pinning versions, still not sure what the loop was for.

MSevey:
have we tested this on a dev server?

MSevey:
> > > also according to [ansible pip docs](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/pip_module.html) we don't need to define loop, we can just define a number of packages we want installed
> > 
> > 
> > the loop was to refactor the packages since they were all using latest. so yes, if you are defining different images for all the packages it doesn't make sense to use the loop.
> 
> Can you clarify? We could have used "name" list even without pinning versions, still not sure what the loop was for.

followed up on discord

kwypchlo:
> have we tested this on a dev server?

I tested it on as-hk-1 when deploying latest tag and it worked just fine

MSevey (approved):

PR #302, opened 2022-04-28
improve apt tasks

kwypchlo:
- removed standalone "apt-get update" tasks in favour of updating apt-get when running a task that installs a package
- replaced "loop" with a "name" list
- increased "cache_valid_time" from 15 minutes to 30 minutes to increase the chance that we will only have to update once since most jobs should finish in 30 minutes (we can increase it to 60 minutes too fwiw)
- replaced "state: present" with "state: latest" to make sure packages are up to date
- unfortunately we can't pin specific versions of packages since linux distro version will have a little bit different package versions (for example different versions of ubuntu have different available tmux versions)


MSevey:
> unfortunately we can't pin specific versions of packages since linux distro version will have a little bit different package versions (for example different versions of ubuntu have different available tmux versions)

We currently have a lock on the linux distros for the portals though. Debian 10 and debian 11. So shouldn't we be able to set the apt package versions?

MSevey:
if we are cleaning this up, we should clean it up all the way, which would include refactoring it into a helper tasks so that the options of `apt` are only defined once and is consistent across plays.

kwypchlo:
> > unfortunately we can't pin specific versions of packages since linux distro version will have a little bit different package versions (for example different versions of ubuntu have different available tmux versions)
> 
> We currently have a lock on the linux distros for the portals though. Debian 10 and debian 11. So shouldn't we be able to set the apt package versions?

we also support ubuntu but unfortunately even debian 10 and 11 have different packages available on apt

MSevey:
> > > unfortunately we can't pin specific versions of packages since linux distro version will have a little bit different package versions (for example different versions of ubuntu have different available tmux versions)
> > 
> > 
> > We currently have a lock on the linux distros for the portals though. Debian 10 and debian 11. So shouldn't we be able to set the apt package versions?
> 
> we also support ubuntu but unfortunately even debian 10 and 11 have different packages available on apt

sad.

well then we should still refactor all these apt calls into a task so that they all are handled the same way in terms of cache updates. That will make updates in the future easier.

playbooks/deploy-machines-setup.yml
```diff
@@ -25,22 +25,16 @@
...
       include_tasks: tasks/check-server-os-compatibility.yml
 
-    # Configure server
-    - name: Run "apt-get update"
-      apt:
-        update_cache: yes
-        cache_valid_time: 900
-      become: True
-
     - name: Install packages
```
MSevey, line 28: We should refactor this out in a helper task that accepts a list of packages. 

then we ensure the `update_cache cache_valid_time state` values are always the same for every call

MSevey (changes requested):

playbooks/deploy-machines-setup.yml
kwypchlo: @MSevey maybe you would like to reconsider your change request, this is pretty good improvement as is and making it a helper task could be resolved in potential follow up

MSevey: yea fine as a follow up

MSevey (approved):

playbooks/deploy-machines-setup.yml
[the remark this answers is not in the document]
kwypchlo: follow up in https://github.com/SkynetLabs/ansible-playbooks/issues/344

firyx (approved):

PR #310, opened 2022-05-16
Add option to override server provider default DNS nameservers

firyx:
# PULL REQUEST

## Overview

Novogara server provider is using local DNS caching resolver `unbound`. It seems it is not configured correctly and doesn't resolve some domains, e.g. `sia-storage.net`.

This PR removes usage of `unbound` and uses custom defined nameservers set via `ansible-private` variable.

## Checklist
 - [x] Changelog file created

### DIFF
--- a/playbooks/group_vars/webportals.yml
+++ b/playbooks/group_vars/webportals.yml
@@ -214,6 +214,18 @@ webportal_common_config_defaults:
 # TODO: default to: False
 webportal_setup_health_checks: True
 
+# Define custom nameservers to be used to optionally override hosting provider
+# defaults.
+# NOTE: Be sure to check with server provider if it ok to just override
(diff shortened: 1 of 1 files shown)


MSevey (approved):

kwypchlo (changes requested):
> This PR removes usage of unbound and uses custom defined nameservers set via ansible-private variable.

I don't think I like this solution, using unbound has its merits and we're removing it even if it's the default that came with debian distro. We should figure out what was wrong and fix that instead of dropping the cache layer completely and adding complexity for portal operators - how would a portal operator even decide which server he needs to overwrite the nameservers on, there's not immediate error or anything happening, the only way we figured out that the dns was not resolving was a user reporting the issue.

PR #315, opened 2022-05-25
Fix errors with handling authenticated portals in the default case

MSevey:
# PULL REQUEST

## Overview
There were some bugs in the API Key flow for the integrations tests. These revolved around a mismatch of a blank `accounts_limit_access` in ansible not knowing that the default is `authenticated`. 

## Example for Visual Changes

Successful `portals-setup-following` and integration test run on my portal seveysky.net

## Checklist
 - [ ] Changelog file created


playbooks/templates/.env.j2
```diff
@@ -176,19 +176,18 @@ SKYNET_ACCOUNTS_LOG_LEVEL={{ webportal_common_config.accounts_log_level }}
 # - 'false': All users incl. anonymous users can access the portal (access to portal is public)
 # REQUIRED: manually defined by user, or default is set in docker-compose file
 ACCOUNTS_LIMIT_ACCESS={{ webportal_common_config.accounts_limit_access | string | lower }}
+{% endif %}
 
-{% if webportal_common_config.accounts_limit_access | string | lower != 'false' %}
```
firyx, line 180: Are `ACCOUNTS_TEST_USER_...` vars required when `ACCOUNTS_LIMIT_ACCESS=false`, i.e. on anonymous portal? I thing they are not required in this case and should be enclosed in `{% if %}` block.

firyx (approved):

playbooks/templates/.env.j2
firyx: .

firyx (changes requested):

playbooks/templates/.env.j2
firyx, line 180: Are `ACCOUNTS_TEST_USER_...` vars required when `ACCOUNTS_LIMIT_ACCESS=false`, i.e. on anonymous portal? I thing they are not required in this case and should be enclosed in `{% if %}` block.

firyx (changes requested):

playbooks/templates/.env.j2
MSevey: yes I think you're right, I think this if statement should stay. The original issue was that this if statement shouldn't be enclosed in the outer if statement because `account_limit_access` can be `undefined` and these values are needed.

I'll update. Good catch.

playbooks/templates/.env.j2
```diff
@@ -176,19 +176,20 @@ SKYNET_ACCOUNTS_LOG_LEVEL={{ webportal_common_config.accounts_log_level }}
 # - 'false': All users incl. anonymous users can access the portal (access to portal is public)
 # REQUIRED: manually defined by user, or default is set in docker-compose file
 ACCOUNTS_LIMIT_ACCESS={{ webportal_common_config.accounts_limit_access | string | lower }}
+{% endif %}
 
 {% if webportal_common_config.accounts_limit_access | string | lower != 'false' %}
```
firyx, line 181:
```suggestion
{% if webportal_common_config.accounts_limit_access | deafult('') | string | lower != 'false' %}
```

I think we need a default here if it is undefined in LastPass.

playbooks/templates/.env.j2
```diff
@@ -176,19 +176,20 @@ SKYNET_ACCOUNTS_LOG_LEVEL={{ webportal_common_config.accounts_log_level }}
...
-# Account Test account for health checks.
+# Account Test account for health checks and integration tests.
 #
-# REQUIRED when ACCOUNTS_LIMIT_ACCESS is not 'false': manually defined by user.
```
firyx, line 183: This comment should stay.

MSevey:
ansible lint action was updated and is pulling in a bunch of new checks.
will open a new PR to address those as to not clutter this PR.

playbooks/templates/.env.j2
```diff
@@ -176,19 +176,20 @@ SKYNET_ACCOUNTS_LOG_LEVEL={{ webportal_common_config.accounts_log_level }}
...
 
-{% if webportal_common_config.accounts_limit_access | string | lower != 'false' %}
-# Account Test account for health checks.
+{% if webportal_common_config.accounts_limit_access | default('') | string | lower != 'false' %}
+# Account Test account for health checks and integration tests.
 #
 # REQUIRED when ACCOUNTS_LIMIT_ACCESS is not 'false': manually defined by user.
 # API key takes precedence over email and password. It is recommended.
-# Email and password are deprecated and will be removed in the future.
+# Email and password are deprecated in use for the health checks but are needed
```
kwypchlo, line 186: I think the comment here was valid. We should be able to remove ACCOUNTS_TEST_USER_EMAIL and ACCOUNTS_TEST_USER_PASSWORD since we don't need to store it once we have the api key in ACCOUNTS_TEST_USER_API_KEY.

MSevey: Some service is still failing for it being required. Need to find that service and fix that.

playbooks/templates/.env.j2
```diff
@@ -176,19 +176,20 @@ SKYNET_ACCOUNTS_LOG_LEVEL={{ webportal_common_config.accounts_log_level }}
 # - 'false': All users incl. anonymous users can access the portal (access to portal is public)
 # REQUIRED: manually defined by user, or default is set in docker-compose file
 ACCOUNTS_LIMIT_ACCESS={{ webportal_common_config.accounts_limit_access | string | lower }}
+{% endif %}
 
-{% if webportal_common_config.accounts_limit_access | string | lower != 'false' %}
-# Account Test account for health checks.
+{% if webportal_common_config.accounts_limit_access | default('') | string | lower != 'false' %}
```
ro-tex, line 181: nit: This `default` should be `authenticated` in order to be consistent with `playbooks/tasks/portal-configs-load.yml`. But the result is the same, therefore this is a `nit` and not ` req`.

playbooks/templates/.env.j2
ro-tex: We should update the health check repo because API keys are usable for its purposes. We should double-check but I believe the only thing it needs is to verify the tier of the user, which it can do by calling `GET /user/limits`, which supports API keys.

ro-tex (approved):

playbooks/templates/.env.j2
```diff
@@ -176,19 +176,20 @@ SKYNET_ACCOUNTS_LOG_LEVEL={{ webportal_common_config.accounts_log_level }}
...
-# Email and password are deprecated and will be removed in the future.
+# Email and password are deprecated in use for the health checks but are needed
+# for generating the API key.
 ACCOUNTS_TEST_USER_API_KEY={{ webportal_server_config.accounts_test_user_api_key }}
 ACCOUNTS_TEST_USER_EMAIL={{ webportal_server_config.accounts_test_user_email }}
 ACCOUNTS_TEST_USER_PASSWORD={{ webportal_server_config.accounts_test_user_password }}
-
-{% endif %}
 {% endif %}
+
```
firyx, line 192: Keep the original formatting, otherwise if `if` condition is false, double empty line is rendered.

playbooks/templates/.env.j2
MSevey: @ro-tex here we are not setting the value of `acconuts_limit_access`, we are checking it. That is why the default is blank, Because we don't need to set it here or in ansible.

playbooks/templates/.env.j2
MSevey: Created a follow up https://linear.app/skynetlabs/issue/SKY-828/remove-test-email-and-test-password-from-env-file

Shouldn't block this PR.

playbooks/templates/.env.j2
MSevey: I don't understand why a space after the if instead of a space before the if matters here. At the end of the day it is one space. If a space is added when the if is removed, there would be two spaces in both cases right?

Also, formatting lint this feels like a `nit` or at least a follow up and shouldn't be a blocking comment.

firyx: This is because conditional blocks in this template are done this way:

```
some-row-1

{if}
optional-row-2

{endif}
some-row-3
```

When condition is true:
```
some row-1

optional-row-2

some-row-3
```

When condition is false and empty row after the block is within if block:
```
some-row-1

some-row-3
```

When condition is false and empty row after the block is outside if block:
```
some-row-1


some-row-3
```

firyx: nit

firyx (approved):

ro-tex (approved):

PR #317, opened 2022-06-02
Fix docker flags for portal_modules. 

MSevey:
…the docker flags are set

# PULL REQUEST

## Overview

The way the `portal_modules` variable from `hosts.ini` was being used didn't properly set the docker flags. 

## Checklist

- [ ] All git commits are signed. (REQUIRED)
- [ ] All new methods or updated methods have clear docstrings.
- [ ] Appropriate documentation updated.
- [ ] Changelog file created.

## Issues Closed

Closes SKY-786

### DIFF
--- a/playbooks/tasks/portal-configs-load.yml
+++ b/playbooks/tasks/portal-configs-load.yml
@@ -100,6 +100,12 @@
       {% endfor %}
   failed_when: mongodb_config_server_preference_result.failed | default(False)
 
+# If portal_modules is defined in hosts.ini, update the values
+- name: Update portal_modules
+  set_fact:
+    webportal_server_config: "{{ webportal_server_config | combine({'portal_modules': portal_modules}, recursive=True) }}"
+  when: portal_modules is defined
+
 - name: Set Accounts flag
   set_fact:
     portal_accounts_on: >-
--- a/playbooks/templates/.env.j2
+++ b/playbooks/templates/.env.j2
@@ -59,7 +59,7 @@ SIA_API_PASSWORD={{ webportal_server_config.sia_api_password }}
 # PORTAL_MODULES=aj
 #
 # REQUIRED: manually defined by user, default is mongo
-PORTAL_MODULES={{ portal_modules | default(webportal_server_config.portal_modules) }}
+PORTAL_MODULES={{ webportal_server_config.portal_modules }}
 
 # Sia wallet password
 #


playbooks/tasks/portal-update-env-value.yml
```diff
@@ -0,0 +1,12 @@
...
+#
+# TODO: Currently this is only set up to update portal_modules but it should be
+# updated to handle any updates to the .env file for deploys when a full
+# regeneration of the .env isn't needed.
+
+- name: Update PORTAL_MODULES
+  ansible.builtin.command:
+    cmd: |
+      sed -i 's/^PORTAL_MODULES=.*/PORTAL_MODULES={{portal_modules}}/' {{ webportal_dir }}/.env
+    warn: False
```
kwypchlo, line 12: let's use a built in module for that
https://docs.ansible.com/ansible/latest/collections/ansible/builtin/lineinfile_module.html

kwypchlo (changes requested):

playbooks/tasks/portal-docker-compose-files-get-wanted.yml
```diff
@@ -5,16 +5,23 @@
...
-    portal_modules: "{{ portal_modules_result.stdout }}"
+- block:
+    - name: Get PORTAL_MODULES from .env file
+      ansible.builtin.shell: ". {{ webportal_dir }}/.env && echo $PORTAL_MODULES"
+      register: portal_modules_result
+
+    - name: Set portal modules var
+      set_fact:
+        portal_modules: "{{ portal_modules_result.stdout }}"
+  when: portal_modules is not defined
```
firyx, line 24: This is not correct. In this place it should be always read from `.env` file.
Why: If we change `portal_modules` in `hosts.ini` without running portals-setup-following some modules can be enabled, but they will be missing config, because `.env` file is not fully generated with newly turned on modules.

So if we change `portal_modules` in either LastPass or in hosts.ini, changes should not take effect until we run portals-setup-following.

playbooks/tasks/portal-update-env-value.yml
firyx, line 12: As mentioned above, if we enable portal modules, we need to regenerate `.env` file with their configs, we shouldn't enable portal modules without running portals-setup-following.

firyx (changes requested):

playbooks/tasks/portal-configs-load.yml
```diff
@@ -100,6 +100,12 @@
       {% endfor %}
   failed_when: mongodb_config_server_preference_result.failed | default(False)
 
+# If portal_modules is defined in hosts.ini, update the values
+- name: Update portal_modules
+  set_fact:
+    webportal_server_config: "{{ webportal_server_config | combine({'portal_modules': portal_modules}, recursive=True) }}"
```
firyx, line 106: What is wanted behavior here?

I thought we wanted this:
- if portal modules are set for specific server in LastPass, we use exactly this (not `combine`)
- otherwise we use value from `hosts.ini`

I other words: portal modules are defined in `hosts.ini` with optional override from server config.
If we use `combine` here we set portal modules to union of the above configs.

firyx (changes requested):

playbooks/tasks/portal-configs-load.yml
MSevey: the goal here is that we are consistent between LastPass and hosts.ini

so i want to override the field. so should I not have `recursive=True` here?

firyx: Looking once more on this.

I think we shouldn't save configs from `hosts.ini` back to LastPass (via `webportal_server_config`).

Current process is (without `hosts.ini` configs):

- load server config(s) from LastPass
- set default values if the vars are not set in webportal_server_config (LastPass)
- save back (missing vars) to server config in LastPass

With `hosts.ini` I think it should be something like:

- load server config(s) from LastPass
- if portal_modules is not defined in `hosts.ini` and portal_modules is not defined in `webportal_server_config` then set default value in `webportal_server_config`
- portal_modules = portal_modules | default(webportal_server_config.portal_modules)
- set default values if the vars are not set in webportal_server_config (but exclude those set in `hosts.ini`)
- save back (missing vars) to server config in LastPass

If we move more and more config vars to `hosts.ini` we could divide `webportal_server_config_defaults` to something like:
- `webportal_server_config_defaults`
- `webportal_server_config_defaults_if_not_in_hosts_ini`

where the first works as usually (sets the default value if missing, to be saved to LastPass)
and the second sets the default value (to be saved to LastPass) only if missing in hosts.ini

firyx (changes requested):

playbooks/tasks/portal-configs-load.yml
MSevey: Summary from Discord Call.

Flows
1) not defined in hosts.ini or LastPass
 - this is error case, fail

2) not defined in hosts.ini, defined in LastPass
 - current state, preserve backwards compatibility
 - load from lastpass into `webportal_server_config.portal_modules`
 - Save back to lastpass 

3) defined in hosts.ini and defined in LastPass
 - override LastPass value on load with hosts.ini value
 - set LastPass value to `Deprecated: moved to hosts.ini` on save

4) defined in hosts.ini and not defined in LastPass
 - new ideal state
 - add check if `webportal_server_config.portal_modules is defined` before setting to `portal_modules`
 - nothing saved back to lastpass 

For `.env` file use `PORTAL_MODULES=portal_modules | default(webportal_server_config.portal_modules)`

firyx (approved):

firyx:
Code on which Karol requested change was removed completely, so merging now.

kwypchlo (dismissed):
let's not merge it in with all those retries in there, those are only necessary because we're running into issues with skyd

ro-tex:
> let's not merge it in with all those retries in there, those are only necessary because we're running into issues with skyd

We might not need the retries if skyd is suddenly perfect but:
1. it's not super likely to be perfect and stay perfect, so the retries are fine
2. if the operation is successful the retries have no impact

@kwypchlo what is the issue with having retries in there?

kwypchlo:
> > let's not merge it in with all those retries in there, those are only necessary because we're running into issues with skyd
> 
> We might not need the retries if skyd is suddenly perfect but:
> 
> 1. it's not super likely to be perfect and stay perfect, so the retries are fine
> 2. if the operation is successful the retries have no impact
> 
> @kwypchlo what is the issue with having retries in there?

Well they do hide issues that might be happening and should not be. We are running those tests in a perfect environment, no external traffic (since server is disabled), virtually no load, directly after fresh start. If we can't have 71 basic api tests pass without an issue in conditions like that and we need retries then something is definitely up. If not for those tests failing without retries we wouldn't be able to pinpoint an issue and eventually fix it.

PR #325, opened 2022-06-10
Add run shell command playbook

firyx:
# PULL REQUEST

## Overview

Adding a playbook to run ad-hoc shell commands against the portals.

## Checklist

- [x] Changelog file created.


playbooks/x-portals-run-shell-command.yml
```diff
@@ -0,0 +1,47 @@
...
+      ansible.builtin.shell: "{{ item }}"
+      loop: "{{ shell_commands }}"
+      register: shell_commands_result
+
+    - name: Log shell command(s) output
+      debug:
+        msg: |
+          command:
+          {{ item.cmd }}
+
```
ro-tex, line 42: Let's fix this lint error.

playbooks/x-portals-run-shell-command.yml
```diff
@@ -0,0 +1,47 @@
+# This playbook expects set `shell_commands` list of commands.
+# E.g. in a file `my-vars/shell-commands.yml` with content:
+#
+# shell_commands:
+#   - "echo abc > abc.txt"
+#   - "docker exec <...>"
```
ro-tex, line 6: Can we wrap these in "```" in order to make it extra clear that this is the example content of the file? I managed to not understand that, somehow. :/

Also, do we need to quotes? I forgot to add them and it was fine.
This if the file that worked for me:
```
shell_commands:
  - cd /home/user/skynet-webportal && nohup ./dc logs -f pinner>logs/pinner.log &
  - curl -L -X POST http://10.10.10.130:4000/sweep
 ```

**Review by ro-tex — DISMISSED** [review]
Awesome! This will be sooo helpful!
And also dangerous. Definitely "handle with care" kind of thing.

**Review by MSevey — COMMENTED** [review]
Is there a way we can combine this with the docker commands? Very similar, potentially the same, if the docker command variable was named better.  My bad 😬

**Comment by firyx** [comment]
> Is there a way we can combine this with the docker commands? Very similar, potentially the same, if the docker command variable was named better. My bad grimacing



> Is there a way we can combine this with the docker commands? Very similar, potentially the same, if the docker command variable was named better. My bad grimacing

Refactored to unite docker/shell commands.
There are now 2 distinct playbooks to run shell commands - safer with and less safer without taking portals out of LB.
Described in updated Readme.

**Comment by firyx** [comment]
@kwypchlo
Once this is merged and changes are pulled, we should update `docker_commands` to `shell_commands` in `deploy-2022-...` configs.

**Comment by firyx** [comment]
Backwards compatibility (no change to existing configs is needed):

- at least one of: older `docker_commands` or newer `shell_commands` must be defined
- when both are defined: Ansible adds those 2 lists and runs all commands

**Comment by firyx** [comment]
> > Backwards compatibility (no change to existing configs is needed):
> > 
> > * at least one of: older `docker_commands` or newer `shell_commands` must be defined
> > * when both are defined: Ansible adds those 2 lists and runs all commands
> 
> hold up, what do you mean at least one has to be defined? defining additional commands should be really optional right?

My bad (incomplete) wording:

- There are 2 playbooks when you explicitly want to run docker/shell commands. For these 2 playbooks you must explicitly set either `docker_commands` or `shell_commands`, otherwise you have something misconfigured - e.g. you forget to load the vars file `-e @my-vars/...`.
- For deployment playbooks `docker_commands` and `shell_commands` are optional.

File: README.md [hunk 3]
```diff
@@ -651,29 +651,73 @@ separate section above).
...
+
+Playbook:
+
+- Runs shell commands defined in `shell_commands` variable and logs outputs.
+
+I.e. commands can be run on live portal - use carefully.
+
+To run:  
+
+`scripts/x-portals-run-shell-commands-without-disable-without-test.sh -e @my-vars/shell-commands.yml --limit eu-fin-1`
```
MSevey, line 713: REQ: I wouldn't commit this. If we, or any dev, wants to do something non-standard like this, they should just have it locally. We as a team could push a branch that team members could utilize, but it shouldn't be in master. 

I would remove the playbook and script from this PR as well.

README.md
```diff
@@ -651,29 +651,73 @@ separate section above).
...
+- Runs shell commands defined in `shell_commands` variable and logs outputs.
+- Runs integration tests and health checks.
+- Enables portals.
+
 To run:  
-`scripts/portals-docker-command.sh -e @my-vars/config.yml --limit eu-fin-1`
 
-The deploy script also supports the docker command execution:
-`scripts/portals-deploy.sh -e @my-vars/config.yml --limit eu-fin-1`
+`scripts/portals-run-shell-commands-with-disable-run-test-enable.sh -e @my-vars/shell-commands.yml --limit eu-fin-1`
```
MSevey, line 699: nit:

```suggestion
`scripts/portals-run-shell-commands.sh -e @my-vars/shell-commands.yml --limit eu-fin-1`
```

With my other comment about not including the option to run these without disabling the LB, this should be simplified.

README.md
```diff
@@ -651,29 +651,73 @@ separate section above).
...
 
 ```yaml
 ---
-docker_commands:
+shell_commands:
   - "docker exec sia siac"
   - "docker exec sia siac renter"
 ```
 
+Wrapping commands into quotes is optional, but might be needed if Ansible/Yaml
```
MSevey, line 683: Good to know :-)

playbooks/tasks/portal-execute-commands.yml
```diff
@@ -1,7 +1,14 @@
...
-- name: Executing Docker Command
-  include_tasks: tasks/portal-docker-command.yml
+# Execute any defined docker/shell command(s).
+# This loops over all the docker/shell commands listed and executes them one at a
+# time. Variable docker_commands is depreciated, shell_commands should be used
+# instead.
+- name: Executing shell command(s)
+  vars:
+    commands: "{{ (docker_commands | default([])) + (shell_commands | default([])) }}"
+  ansible.builtin.shell: "{{ item }}"
```
MSevey, line 9: REQ: Shouldn't this be `tasks/shell-commands-run-subtask.yml`?

MSevey (changes requested):

playbooks/tasks/portal-execute-commands.yml
firyx: Currently during deploys we don't log commands' outputs (unless the command fails), so I didn't want to spam deploy logs as the subtask would. Also I didn't want to introduce an unnecessary complexity adding a new "no log" switch to the shell commands task/subtask so I kept here a simple loop/shell task.

README.md
firyx: Removed script and playbook from the PR/master.
Will create `internal` branch which keeps the playbook.

playbooks/tasks/portal-execute-commands.yml
firyx: Currently running docker/shell commands doesn't spam deploy logs with commands outputs unless the command fails.

I can make it reuse the subtask (prioritizing DRY principle) and to keep not spamming logs we would need to add a new flag to not log to ansible logs during deploys, but to log when explicit run commands playbook (or internal playbook) is run which will add another complexity to playbooks.

Or we can keep the playbooks cleaner (prioritizing KISS principle) with this simple loop/shell task.

LMK if you still want to prioritize DRY over KISS?

interstella-5555 (dismissed):

playbooks/tasks/portal-execute-commands.yml
firyx: Decided on discord:
- DRY
- keep displaying command output logs

MSevey (dismissed):
Looks good. Like turn my one follow up comment into an issue and plan it into a future cycle.

README.md
```diff
@@ -651,29 +651,59 @@ separate section above).
...
+#### Disable from a load balancer, run shell commands, test and enable portals
+
+See above for `shell_commands` configuration.
+
+Playbook:
+
+- Disables health check.
+- Runs shell commands defined in `shell_commands` variable and logs outputs.
+- Runs integration tests and health checks.
+- Enables portals.
```
ro-tex, line 695: How do we run a command without disabling health checks?

It is surprising to me that running a shell command will pull a sever from the LB. That seems to be a responsibility which should lie with the operator - they are aware of what type of command they are running and whether that requires the server to be pulled first.

I suggest we make that configurable via a parameter and switch it off by default.

README.md
```diff
@@ -651,29 +651,59 @@ separate section above).
...
+  `my-vars/shell-commands.yml`
+
+Depreciation note:
+
+Previously there was `docker_commands` variable with the similar functionality,
+its usage is now depreciated, but for backwards compatibility it works as it
+used to work.
+
+If both `docker_commands` and `shell_commands` are used, commands in both of
+these lists are run (commands from these two lists are added together).
```
ro-tex, line 672: Do these variables refer to section names in the passed `-e @my-vars/something.yml`? I guess so from the context and examples. It might be useful to state that explicitly for non-savvy ansible users like me who are just doing they best to get stuff done without really knowing a lot about ansible.

playbooks/portals-run-shell-commands.yml
```diff
@@ -12,14 +14,15 @@
...
+    portal_action: "portal-run-shell-commands"
 
   tasks:
     # Check '--limit' is used. This is to protect against taking the entire
     # fleet down
     - name: Fail if you are targeting all dev and prod webportals
       include_tasks: tasks/host-limit-check.yml
 
-    # Skip portals on takedown as they are already out of the LB
+    # Skip portals on takedown
```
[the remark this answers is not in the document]
ro-tex: I second this. I am not sure this should be a F/U or a blocker - not being able to run shell commands against portals out of the LB is very limiting. We are running these shell commands against a manually defined list of servers, it shouldn't matter whether they are in or out of the LB.

As I stated elsewhere, the entire concept of in/out of LB should be a responsibility of the operator and the scripts should provide an command line option for that, at best.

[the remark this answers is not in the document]
kwypchlo: Takedown process will be refactored in https://linear.app/skynetlabs/issue/SKY-1251/takedown-servers-should-run-all-services-but-redeploy-with-new-env

README.md
firyx: I think that your question is covered in readme:

- `shell_commands` variable ...
- ... can be defined either in `my-vars/portal_versions.yml` (e.g. for
  `portals-deploy` playbook with other variables) or in a separate file e.g.
  `my-vars/shell-commands.yml`

There is an example code, how to set it inside yaml file
and example code how to run it from shell.

playbooks/portals-run-shell-commands.yml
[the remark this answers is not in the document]
firyx: I have updated the PR so that shell commands run also on takedown servers.

interstella-5555 (approved):

MSevey (approved):

README.md
MSevey: > It is surprising to me that running a shell command will pull a sever from the LB. That seems to be a responsibility which should lie with the operator - they are aware of what type of command they are running and whether that requires the server to be pulled first.

I think that makes sense when doing things manually, but i disagree with ansible playbooks.  The pattern we have enforced is that playbooks handle everything needed safely. Making developers have to know when to manually do things along side of playbooks is an anti-pattern IMO.

ro-tex: Something I've used this playbook for multiple times already is to kick off Pinner sweeps on large groups of servers. Another use case I can easily imagine is checking the free disk space on a bunch of servers. Neither of these require servers to be taken out of the LB. Moreover, I would argue that the majority of operations we are ever going to use this playbook for fall in the same category - minor operations, which don't affect the state of the server.

Given the above, having the playbook take servers out of the LB has two drawbacks:
 - a distracted developer can bring down most of the cluster by running a simple `df -h` against it
 - any such operation will take exorbitant amounts of time, despite the only purpose of this playbook being to save us time

If we have the same distracted dev from before and they run something that should bring a server out of the LB and they forget to do that, I would expect the health checks of that server to bring it out the next time they run. That's definitely an undesirable situation but it seems better than a complete cluster meltdown, which would happen if we accidentally remove the majority of the servers and the others die from too much traffic.

**Suggestion**:
In order to make this both usable and safe, I suggest we add a *required* parameter which will define whether the server will be taken out of the LB or not. If the parameter is undefined, the playbook fails to run.

ro-tex (changes requested):
We can't merge this before coming up with a satisfactory solution to the "out of LB" problem. Otherwise we're creating a major footgun.

interstella-5555:
not on a current roadmap

PR #327, opened 2022-06-14
replace siac with skyc

kwypchlo:
- replace `docker exec sia siac` with `docker exec sia skyc`
- replace other siac references for consistency
- this is backwards compatible because we've been aliasing `siac` with `skyc` since forever in our docker image anyway


firyx (approved):

MSevey (changes requested):
We need to make this backwards compatible on the servers before merging.

kwypchlo:
due to severe compatibility issues with what seemed like a small change, this has been dropped from the current roadmap

PR #330, opened 2022-06-22
Fil/sky 1060 wait for consensus sync before full setup

firyx:
# PULL REQUEST

## Overview

Current playbooks wait in the following order:
- wait for Sia finished full setup
- wait for Sia consensus sync

This PR swaps order of waits so that portal operator is more accurately informed what happened when the wait fails.

## Checklist

- [ ] Changelog file created.

### DIFF
--- a/playbooks/group_vars/webportals.yml
+++ b/playbooks/group_vars/webportals.yml
@@ -275,8 +275,10 @@ portal_health_check_disable_delay_secs: 0
 # Disable portal when free disk space is lower then x GiB (60 GiB recommended)
 portal_health_check_disable_free_disk_space_limit_bytes: "{{ 60e+9 | int }}"
 
-# Waiting for sia full setup and /daemon/ready
-sia_full_setup_timeout_secs: 900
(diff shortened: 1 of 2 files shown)


MSevey (changes requested):
skyd will complete a full set up on a fresh server before the consensus is sync. So this change is an anti-pattern.

What was the reason behind this change?

firyx:
>

firyx:
> skyd will complete a full set up on a fresh server before the consensus is sync. So this change is an anti-pattern.
> 
> What was the reason behind this change?

I'm observing it otherwise:
- sia full setup doesn't finish until consensus is synced

MSevey:
> I'm observing it otherwise:

Is this on repeated runs of the setup playbook?

if so then we can figure out what is causing it to not finish the setup. 

a syncing consensus on its own is not the blocking action. So we shouldn't switch these. We should figure out which action is blocking the setup due to the consensus syncing and see if that needs to be blocking.

MSevey (changes requested):
Pending longer discussion on root issue.

PR #331, opened 2022-06-26
Add ACCOUNTS_REDIRECT_* env variables in .env.j2 template

meeh0w:
## Overview
* Adds & describes env variables added by https://github.com/SkynetLabs/webportal-nginx/pull/17

- [ ] All new methods or updated methods have clear docstrings.
- [ ] Changelog file created.

### DIFF
--- a/playbooks/templates/.env.j2
+++ b/playbooks/templates/.env.j2
@@ -125,6 +125,16 @@ AIRTABLE_BASE={{ webportal_common_config.airtable_base | default('') }}
 AIRTABLE_TABLE={{ webportal_common_config.airtable_table | default('') }}
 AIRTABLE_FIELD={{ webportal_common_config.airtable_field | default('') }}
 
+# Used to redirect traffic coming to account.<portal_domain> subdomain.
+# Mentioned redirect will be active in two scenarios:
+#   - whenever ACCOUNTS_REDIRECT_URL is specified - users will be redirected to configured URL
+#   - when accounts module is disabled - if this variable is not configured, users will be 
+#     redirected to PORTAL_DOMAIN by default.
+ACCOUNTS_REDIRECT_URL={{ webportal_common_config.accounts_redirect_url | default('') }}
+# The redirect above will use HTTP status code 302 by default (temporary redirect).
+# You're free to configure it to meet your needs by setting the variable below.
+ACCOUNTS_REDIRECT_STATUS_CODE={{ webportal_common_config.accounts_redirect_status_code | default('') }}
+
 ###############################################################################
 # MongoDB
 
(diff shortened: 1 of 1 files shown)


playbooks/templates/.env.j2
```diff
@@ -125,6 +125,16 @@ AIRTABLE_BASE={{ webportal_common_config.airtable_base | default('') }}
 AIRTABLE_TABLE={{ webportal_common_config.airtable_table | default('') }}
 AIRTABLE_FIELD={{ webportal_common_config.airtable_field | default('') }}
 
+# Used to redirect traffic coming to account.<portal_domain> subdomain.
+# Mentioned redirect will be active in two scenarios:
+#   - whenever ACCOUNTS_REDIRECT_URL is specified - users will be redirected to configured URL
+#   - when accounts module is disabled - if this variable is not configured, users will be 
+#     redirected to PORTAL_DOMAIN by default.
+ACCOUNTS_REDIRECT_URL={{ webportal_common_config.accounts_redirect_url | default('') }}
```
meeh0w, line 133: Now to have it in production for siasky.net I'll need to update `Shared-Ansible/cluster-prod.yml`, right?

If so, does `Shared-Ansible-Pro/cluster-prod.yml` "inherit" from that file? Meaning, should I also update this file for SkynetPro and override with empty values, or is this not needed?

firyx: `accounts_redirect_url` will be just plain URL, no secrets in query string ,right?
which portals should be affected: siasky.net & skynetfree & skynetpro?

generally we are now trying to keep just secrets in LastPass (and later in Vault),
non-secrets we should keep in https://github.com/SkynetLabs/ansible-private
esp. here: https://github.com/SkynetLabs/ansible-privatall the others e/tree/master/custom-vars

currently there is `common.yml` loaded for all portals incl. dev(s), siasky.net, free, pro
then:
- skynetfree loads `skynetfree.yml`
- skynetpro lodds `skynetpro.yml`
- all the others load `skynet-default.yml`
loading is controlled in <https://github.com/SkynetLabs/ansible-private/blob/master/inventory/hosts.ini>

so if we need to set it for siasky.net & free & pro
and not set it for others like dev portals

we should create `siaskynet.yml` out of `skynet-default.yml`
update `hosts.ini` loading `siaskynet.yml` under `[webportals_prod:vars]`

and add variable to the 3 yamls above:
`accounts_redirect_url: https:/...`

then `.env.j2` should be updated just to `ACCOUNTS_REDIRECT_URL={{ waccounts_redirect_url | default('') }}`

firyx (changes requested):
see my comment to what is now preffered location of non-secret config variables

playbooks/templates/.env.j2
meeh0w: Thanks for all the tips 👍 I'm on it.

firyx (approved):
I'm approving this PR,
it sets default values,
but note that to make the change active on `siasky.net` portals,
accounts redirect var need to be prepared in `ansible-private` repo
and portal-setup-following playbook must be executed on all `siasky.net` servers.

meeh0w:
@firyx yup, I have a branch ready for ansible-private ([here](https://github.com/SkynetLabs/ansible-private/compare/master...sky-797)), just waiting for all other dependencies to merged (I can't make a draft PR)

MSevey (approved):

PR #334, opened 2022-07-08
support deny_public_access as a replacement for takedown

kwypchlo:
Because rewriting "takedown" feature to be backwards compatible would be almost impossible, I introduced new variable that you can pass in hosts.ini that replaces takedown functionality. Takedown should become deprecated after this is merged.

Example usage in hosts.ini:

```ini
[webportals_siasky]
...
as-hk-1 ansible_host=as-hk-1.siasky.net
as-sp-1 ansible_host=as-sp-1.siasky.net
as-sp-2 ansible_host=as-sp-2.siasky.net deny_public_access=true
eu-fin-1 ansible_host=eu-fin-1.siasky.net
...
```

Once this is merged, we should rewrite our infrastructure's hosts.ini takedown into deny_public_access.

### DIFF
--- a/playbooks/templates/.env.j2
+++ b/playbooks/templates/.env.j2
@@ -61,6 +61,14 @@ SIA_API_PASSWORD={{ webportal_server_config.sia_api_password }}
 # REQUIRED: manually defined by user, default is mongo
 PORTAL_MODULES={{ portal_modules | default(webportal_server_config.portal_modules) }}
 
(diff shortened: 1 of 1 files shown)


kwypchlo:
@firyx tested the setup-following on siasky.xyz and it works great :)

peterjan (approved):

firyx (approved):

PR #342, opened 2022-07-27
Adjust ansible to accept skyd docker image

kwypchlo:
**Case: `docker-compose.yml` uses dockerfile build and `portal_skyd_version` is not set**
- [x] expected result: `docker-compose.override.yml` does not have entry for sia service as it's going to use the default portal-latest from `docker-compose.yml`

**Case: `docker-compose.yml` uses dockerfile build and `portal_skyd_version` is set to v1.6.0**
- [x] expected result: `docker-compose.override.yml` overrides build branch argument with "v1.6.0"

**Case: `docker-compose.yml` uses image and `portal_skyd_version` is set not set**
- [x] expected result: `docker-compose.override.yml` does not have entry for sia service as it's going to use the default image from `docker-compose.yml`

**Case: `docker-compose.yml` uses image and `portal_skyd_version` is set to v1.6.0**
- [x] expected result: `docker-compose.override.yml` overrides build property defining path to `Dockerfile` from docker-skyd repo (scratch image) with branch argument set to "v1.6.0" (when both `image` and `build` props are defined, docker-compose will build the image locally from the dockerfile defined in `build`)

**Case: `docker_image_overrides` includes custom image for skyd and `docker-compose.yml` uses dockerfile build**
- [x] expected result: this should probably error out, it's not possible to override the build with image

**Case: `docker_image_overrides` includes custom image "foo/bar:1.5.5" for skyd and `docker-compose.yml` uses image and `portal_skyd_version` is set not set**
- [x] expected result: `docker-compose.override.yml` overrides image with "foo/bar:1.5.5"

**Case: `docker_image_overrides` includes custom image "foo/bar:1.5.5" for skyd and `docker-compose.yml` uses image and `portal_skyd_version` is set to v1.6.0**
- [x] expected result: this should probably error out


playbooks/portals-get-versions.yml
```diff
@@ -22,7 +22,6 @@
           - git commit {{ portal_repo_commit }}
 
           - portal_repo_version: {{ portal_repo_version }}
-          - portal_skyd_version: {{ portal_skyd_version_used }}
```
firyx, line 25: I would not remove reporting `skyd` version, but would make it analogous to reporting `accounts` version.
`accounts` version reports either selected branch or image.

firyx (changes requested):

playbooks/portals-get-versions.yml
kwypchlo: I added back support for "portal_skyd_version_used" in bd711e6647fc2fbd4032ca218176687e003960b8 but I am not sure we should have that since we wanted to drop "sia" and "accounts" special handling across ansible but maybe it's not time to do that right now.

Tested on xyz:

Case 1: no override, docker compose uses old branch model
Screenshot 2022-08-05 at 15 06 55

Case 2: no override, docker compose uses docker image
Screenshot 2022-08-05 at 15 07 02

Case 3: override skyd with custom branch
Screenshot 2022-08-05 at 15 07 10

Case 4: override skyd with custom image
Screenshot 2022-08-05 at 15 11 18

MSevey (approved):

firyx (approved):

PR #343, opened 2022-07-27
Pull `cypress` image

peterjan:
# PULL REQUEST

## Overview

This PR extends the `portal-role-task.yml` file with an extra step that pulls the `cypress` image. This avoids having to pull it on request when the abuse scanner needs it, ensuring a smooth parse of the abuse email rather than hanging for x seconds waiting on the image being pulled.

It also adds support for the `ABUSE_TMP_DIR` env. variable.

## Example for Visual Changes

N/A

## Checklist

- [ ] Appropriate documentation updated.
- [ ] Changelog file created.

## Issues Closed

Closes SKY-730

### DIFF
--- a/playbooks/tasks/portal-role-task.yml
+++ b/playbooks/tasks/portal-role-task.yml
@@ -365,3 +365,10 @@
       command: crontab -u {{ webportal_user }} {{ webportal_cron_file }}
 
   when: webportal_setup_health_checks
+
+# Pull cypress image to avoid lazily pulling it on the fly when the abuse
+# scanner needs it
+- name: Pull cypress docker image
+  community.docker.docker_image:
+    name: "{{ cypress_docker_image }}"
+    source: pull
--- a/playbooks/templates/.env.j2
+++ b/playbooks/templates/.env.j2
@@ -293,11 +293,6 @@ ABUSE_PORTAL_URL={{ webportal_common_config.abuse_portal_url }}
 # REQUIRED: Hardcoded, not configurable
 ABUSE_SPONSOR="Abuse Scanner"
 
-# Temporary directory used by the abuse scanner
-#
-# OPTIONAL: manually defined by user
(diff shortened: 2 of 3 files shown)


playbooks/templates/.env.j2
```diff
@@ -293,6 +293,11 @@ ABUSE_PORTAL_URL={{ webportal_common_config.abuse_portal_url }}
 # REQUIRED: Hardcoded, not configurable
 ABUSE_SPONSOR="Abuse Scanner"
 
+# Temporary directory used by the abuse scanner
+#
+# OPTIONAL: manually defined by user
+ABUSE_TMP_DIR={{ webportal_common_config.abuse_tmp_dir | default('/tmp/abuse-scanner') }}
```
MSevey, line 299: What is this used for?

A tmp directory doesn't feel like a needed environment variable.  If it is just used during setup then we can just have an ansible variable.

playbooks/tasks/portal-role-task.yml
```diff
@@ -365,3 +365,10 @@
       command: crontab -u {{ webportal_user }} {{ webportal_cron_file }}
 
   when: webportal_setup_health_checks
+
+# Pull cypress image to avoid lazily pulling it on the fly when the abuse
+# scanner needs it
+- name: Pull cypress docker image
```
MSevey, line 371: Why is this here and not part of the abuse scanner?

peterjan: `cypress` image is ran from the container on the host itself, so it is preferred to have the cypress image be available on all servers instead of pulling it in on the fly

playbooks/templates/.env.j2
peterjan: I can throw this out. I thought we usually add an optional variable for every environment variable we use from the environment so it's configurable should portal operators want to deviate from the default value.

MSevey (approved):

firyx (approved):

PR #353, opened 2022-08-12
replace community.docker.docker_compose with dc script usage

kwypchlo:
Replaces "community.docker.docker_compose" with calls to "dc" script.

This pull request should be followed up with a cleanup to remove all steps that are made redundant after dropping accessing docker through "community.docker.docker_compose" - most prominently, all the steps that check for what docker compose files should be included and what modules should be enabled, that's managed completely through "dc".


firyx (approved):

playbooks/tasks/portal-role-task-mongo-check-docker-service.yml
```diff
@@ -80,7 +76,7 @@
   vars:
     mongodb_shell:
       eval: "db.serverStatus()"
-      until: "'\"ok\" : 1' in mongo_shell_result.transformed_output"
+      until: '''"ok" : 1'' in mongo_shell_result.transformed_output'
```
MSevey, line 79: same

playbooks/tasks/portal-docker-services-start.yml
```diff
@@ -50,25 +50,35 @@
...
-    state: present
-    # TODO: Build in parallel, when Ansible support is added
+- name: Pull docker images
+  ansible.builtin.command: "chdir={{ webportal_dir }} ./dc pull"
   become: True
-  become_user: "user"
-  register: docker_compose_start_result
+  become_user: "{{ webportal_user }}"
+
+- name: Build docker images
```
MSevey, line 58: Is this a no-op if the images are pulled from docker hub?

kwypchlo: yes but will be used when deploying a custom branch (for example for skyd)

playbooks/tasks/portal-role-task-mongo-check-docker-service.yml
```diff
@@ -40,7 +36,7 @@
   when: not mongo_service_ok
 
 - name: Wait for mongo container initialization before checking status
-  ansible.builtin.shell: "docker logs mongo | grep '\"Waiting for connections\"'"
+  ansible.builtin.shell: 'docker logs mongo | grep ''"Waiting for connections"'''
```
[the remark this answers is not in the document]
kwypchlo: this was automatically transformed by prettier in my vscode, I tested this playbook so if there would be any issue it would not pass successfully I guess

playbooks/tasks/portal-role-task-mongo-check-docker-service.yml
kwypchlo: this was automatically transformed by prettier in my vscode, I tested this playbook so if there would be any issue it would not pass successfully I guess

playbooks/tasks/portal-role-task-mongo-check-docker-service.yml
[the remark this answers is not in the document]
MSevey: this command has a when condition which means that your playbook run might not have trigger this.  We have had a number of issues in the past due to specific quotation requirements. Please revert this.

playbooks/tasks/portal-role-task-mongo-check-docker-service.yml
MSevey: same

MSevey (changes requested):
quotation changes need to be reverted.

playbooks/tasks/portal-role-task-mongo-check-docker-service.yml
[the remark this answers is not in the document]
kwypchlo:
```suggestion
  ansible.builtin.shell: "docker logs mongo | grep '\"Waiting for connections\"'"
```

playbooks/tasks/portal-docker-services-start.yml
```diff
@@ -50,20 +50,30 @@
...
+  ansible.builtin.command: "chdir={{ webportal_dir }} ./dc build --no-cache"
```
ro-tex, line 59: Am I reading this wrong or are we pulling and rebuilding the service with no cache each time we want to start it?

kwypchlo: Yes we are pulling images every time we start - images are cached though so its skipped if the versions are already pulled.

Yes we are rebuilding without cache every time we start but currently all our services are served from images so this step is skipped completely, it's only relevant when we override an image with link to Dockerfile and we want to make sure that is rebuilt without cache every time because the args for those are branch names and we want to pull the latest branch every time, otherwise it would get the stale version from cache - we're using that sometimes for debugging purposes.

Just to be clear, both these settings were already set up like that with ansible docker-compose plugin, this is merely transforming it into usage of dc instead.

ro-tex: This is bad. We shouldn't be doing it. Not using the docker images is a valid approach and the portal operator shouldn't be punished by a massive delay each time they restart their stack. Let's remove the `--no-cache`, so at least we use the cache and don't incur the delay on each restart.

kwypchlo: We had this same exact discussion 3 months ago and the only way to redeploy latest changes if operator is targeting a branch is to use no-cache option, see https://github.com/SkynetLabs/ansible-playbooks/pull/328

ro-tex (approved):

MSevey (approved):

PR #369, opened 2022-08-29
Add check that hosts.ini was updated for playbooks v2

firyx:
# PULL REQUEST

## Overview

Playbooks v2 require the updated `hosts.ini` file in `ansible-private` repository.
This PR adds a check that `hosts.ini` was updated.

## Checklist

- [ ] Changelog file created.

### DIFF
--- a/playbooks/portals-block-unblock-incoming-traffic.yml
+++ b/playbooks/portals-block-unblock-incoming-traffic.yml
@@ -19,6 +19,9 @@
     - name: Fail if you are targeting all dev and prod webportals
       include_tasks: tasks/host-limit-check.yml
 
+    - name: Include checking hosts.ini compatibility with playbooks v2
+      include_tasks: tasks/secrets-storage-check-hosts-ini-update.yml
+
     - name: Include loading user credentials from secrets storage
       include_tasks: tasks/secrets-storage-common-load-user-credentials.yml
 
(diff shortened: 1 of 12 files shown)


playbooks/portals-block-unblock-incoming-traffic.yml
```diff
@@ -19,6 +19,9 @@
     - name: Fail if you are targeting all dev and prod webportals
       include_tasks: tasks/host-limit-check.yml
 
+    - name: Include checking hosts.ini compatibility with playbooks v2
+      include_tasks: tasks/secrets-storage-check-hosts-ini-update.yml
+
     - name: Include loading user credentials from secrets storage
       include_tasks: tasks/secrets-storage-common-load-user-credentials.yml
```
MSevey, line 26: Should we just put it in here? Since this is largely ensuring that the credentials don't get corrupted?

firyx-lab: There are approx. 3 places in the playbooks where we load or save credentials,
but we can't run the host.ini check (we expect the ansible_become_pass and hash already set).
So the current way was the most simple implementation.

If we move it down one level to credentials loading and saving tasks,
we need to:
- introduce a new flag `skip_checking_hosts_ini_update`
- set this flag for these exception places when we have to skip the check
- run the check conditionally based on the flag | default(do not skip)
- move the check task one level down to the credentials saving and loading tasks

let me know if we go the above way or keep it as is?

PR #380, opened 2022-09-08
Update Pinner env vars

firyx:
# PULL REQUEST

## Overview

Update Pinner vars.

## Checklist

- [ ] Changelog file created.

### DIFF
--- a/playbooks/templates/.env.j2
+++ b/playbooks/templates/.env.j2
@@ -312,10 +312,13 @@ NCMEC_REPORTER_EMAIL={{ webportal_common_config.ncmec_reporter_email }}
 
 {% endif %}
 {% endif %}
-
 {% if portal_pinner_on %}
-# OPTIONAL:
-PINNER_LOG_LEVEL={{ webportal_common_config.pinner_log_level | default('info')}}
+###############################################################################
+# (optional) Pinner
+
+PINNER_LOG_FILE={{ pinner_log_file | default('/logs/pinner.log') }}
+PINNER_LOG_LEVEL={{ pinner_log_level | default('info') }}
+
 {% endif %}
 # TODO: Add serverlist script vars
 # TODO: Once serverlist vars are added to this template update


ro-tex (approved):

playbooks/templates/.env.j2
```diff
@@ -312,10 +312,13 @@ NCMEC_REPORTER_EMAIL={{ webportal_common_config.ncmec_reporter_email }}
...
 {% endif %}
 {% endif %}
-
 {% if portal_pinner_on %}
-# OPTIONAL:
-PINNER_LOG_LEVEL={{ webportal_common_config.pinner_log_level | default('info')}}
+###############################################################################
+# (optional) Pinner
+
+PINNER_LOG_FILE={{ pinner_log_file | default('/logs/pinner.log') }}
```
kwypchlo, line 319: that should not be customisable path, pinner should define the directory where it stores logs and we should mount that directory to /home/user/skynet-webportal/docker/data/pinner/logs in docker-compose file

kwypchlo (changes requested):

playbooks/templates/.env.j2
ro-tex: I don't see a reason for the log file name to not be customisable. Maybe I want a timestamp on it or something?
And I don't see a reason for this request to be a requirement (REQ) blocking the PR.

kwypchlo: > Like this https://github.com/SkynetLabs/skynet-webportal/blob/master/docker-compose.pinner.yml#L20 ?

Yes, my concern is only that the file name shouldn't be an absolute path but a file relative to /logs directory that we've configured as mounted. Otherwise we're going to end up with logs not showing up and persisted in mounted directory silently.

I do feel like this is a req though, there is no point in pushing code that we know is not valid from the beginning.

On the other hand, I am still not sure why this is a configurable option. Example with timestamp is not really something I can accept, you probably want to logrotate that file instead of adding a timestamp manually with env variable that requires full service restart to pick up the change. Is there a real use case you had that made you add this variable?

kwypchlo: On the second thought, I'm not going to hold this minor pull request, although I'm not too happy about that, a follow up issue will be good enough.

kwypchlo (approved):

playbooks/templates/.env.j2
MSevey: REQ: I don't think the default should be defined here. The pinner service itself doesn't have a default path defined yet.  https://github.com/SkynetLabs/pinner/blob/main/conf/configuration.go#L26

I think it is fine for being able to define a custom logfile path, but I think the architecture pattern here isn't what we want. The pinner service should have the default and can also have the documentation around the path being relative or absolute.  Ansible should then follow that design. 

As the team is splitting and dissolving I think it is very important that we aren't adding hidden complexities and content knowledge is places like this. 

```suggestion
{% if pinner_log_file != "" %}
PINNER_LOG_FILE={{ pinner_log_file }}
{% endif %}
```

MSevey (changes requested):

playbooks/templates/.env.j2
MSevey: cc @ro-tex can you add the defaults to the pinner service itself and provide any documentation necessary about the path being absolute vs relative?

ro-tex: PR: https://github.com/SkynetLabs/pinner/pull/63

TL;DR:
* if `PINNER_LOG_FILE` is empty or not defined, `pinner` won't log to disk
* if `PINNER_LOG_FILE` is defined, it will be a subpath of the container-local `/logs` directory, which is mapped on our servers to `/home/user/skynet-webportal/docker/data/pinner/logs`.

kwypchlo (approved):

PR #385, opened 2022-09-15
new logic to wait until all registry access, download and uploads are finished on skyd

kwypchlo:
- new logic on waiting for skyd renter to be idle based on the memory
- if wait time is set to 0 (on dev servers), don't even check if renter is busy (skipped)

@ChrisSchinnerl /renter output is 15M++ in size and it takes around 15s to parse it as json - if we could only pull the memory part of the renter output that would be great

tested on eu-fin-9 and dev2 in different conditions

note: please squash on merge :)


playbooks/tasks/portal-renter-wait-until-idle.yml
```diff
@@ -0,0 +1,69 @@
...
+# Check if sia container is running.
+- name: Check sia container is running
+  community.docker.docker_container_info:
+    name: sia
+  register: sia_docker_container_result
+
+# Get the ip address for the sia container
+- name: Get sia service IP inside docker compose network
+  command: docker inspect -f "{% raw %}{{ range.NetworkSettings.Networks }}{{ .IPAddress }}{{ end }}{% endraw %}" sia
+  register: sia_ip_result
```
firyx, line 15: I think this will fail portal setup playbook on a fresh server.
I would add `ignore_errors: True`.

playbooks/tasks/portal-renter-wait-until-idle.yml
```diff
@@ -0,0 +1,69 @@
...
+
+- block:
+    - name: Set renter wait start time
+      set_fact:
+        renter_wait_start_time: "{{ (lookup('pipe', 'date +%s') | int) if renter_wait_start_time is not defined else renter_wait_start_time }}"
```
firyx, line 20: IIRC `lookup('pipe')` runs on ansible control machine, not on host, times can be desynced.
I would use `ansible.builtin.command` and then `set_fact`.

playbooks/tasks/portal-renter-wait-until-idle.yml
```diff
@@ -0,0 +1,69 @@
...
+      when:
+        - renter_get_result.rc == 0 and renter_get_result.stderr == ""
+        - renter_busy
+  rescue:
+    - name: "Check if ansible should still wait for renter to get idle"
+      set_fact:
+        renter_busy_should_retry: "{{ (lookup('pipe', 'date +%s') | int) - (renter_wait_start_time | int) < renter_busy_check_wait_secs }}"
+
+    - name: "Retry renter busy check"
+      include_tasks: tasks/portal-renter-wait-until-idle.yml
```
firyx, line 60: We could simplify it with:

```yaml
until:
delay:
retries:
```

see examples in the current code

firyx (changes requested):

playbooks/tasks/portal-renter-wait-until-idle.yml
kwypchlo: it doesn't matter if the time is not in sync with the machine, we need only to calculate delta since we want to make sure we wait up to 300s

playbooks/tasks/portal-renter-wait-until-idle-subtask.yml
```diff
@@ -0,0 +1,83 @@
...
+  when: renter_get_result.rc == 0 and renter_get_result.stderr == ""
+
+- name: "Check if ansible should still wait for renter to get idle"
+  set_fact:
+    renter_busy_should_retry: >-
+      {{
+        (lookup('pipe', 'date +%s') | int) - (renter_wait_start_time | int) < renter_busy_check_wait_secs
+        and
+        renter_busy | default(True)
+      }}
```
kwypchlo, line 79: we should probably not retry if renter_get_result failed, if we fail to fetch renter info we might as well just kill skyd

firyx: discussed on discord, let's keep it for now

PR #387, opened 2022-09-20
setup logrotate configuration for webportal services

kwypchlo:
- removes all logrotate.d configurations prefixed with "skynet-webportal-" (prune old ones before copying over new ones)
- copy over all logrotate config files from skynet-webportal repo (ignore errors if running on old version of skynet-webportal that does not have those defined yet)

### DIFF
--- a/playbooks/tasks/portal-role-task.yml
+++ b/playbooks/tasks/portal-role-task.yml
@@ -366,7 +366,7 @@
     - name: Remove all skynet-webportal logrotate configs
       command: rm -f /etc/logrotate.d/skynet-webportal-*
       become: True
-      
+
     - name: Copy over new skynet-webportal logrotate configs
       command: cp {{ webportal_logrotated_dir }}/* /etc/logrotate.d/
       become: True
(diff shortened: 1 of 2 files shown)


playbooks/tasks/portal-role-task.yml
```diff
@@ -361,6 +361,15 @@
 
   when: webportal_setup_health_checks
 
+# Setup logrotate for webportal logs
+- block:
+    - name: Remove all skynet-webportal logrotate configs
+      command: rm -f /etc/logrotate.d/skynet-webportal-*
+
```
firyx, line 368: We need root/sudo for his:

```suggestion
      become: True
      
```

playbooks/tasks/portal-role-task.yml
```diff
@@ -361,6 +361,15 @@
...
+    - name: Copy over new skynet-webportal logrotate configs
+      command: cp {{ webportal_logrotated_dir }}/* /etc/logrotate.d/
```
firyx, line 370:
```suggestion
      command: cp {{ webportal_logrotated_dir }}/* /etc/logrotate.d/
      become: True
```

firyx (approved):

firyx (changes requested):
We need root permissions.

firyx (approved):

PR #391, opened 2022-09-24
delete old logrotate files before copying new ones

kwypchlo:
Turns out command `rm -f /etc/logrotate.d/skynet-webportal-*` did not work as expected (asterisk was evaluated against local shell) and no files were deleted and the subsequent command that copies files does not overwrite files so once files were copied, they were not updated. This pull request changes the way we delete existing logrotate config files from host to be more reliable.

✅ Tested on dev servers and already deployed with on pro/free servers.

### DIFF
--- a/playbooks/tasks/portal-role-task.yml
+++ b/playbooks/tasks/portal-role-task.yml
@@ -373,7 +373,7 @@
       file:
         state: absent
         path: "{{ item.path }}"
-      with_items: "{{ existing_logrotate.files }}"
+      loop: "{{ existing_logrotate.files }}"
       become: True
 
     - name: Check webportal logrotate config directory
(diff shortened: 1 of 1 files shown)


playbooks/tasks/portal-role-task.yml
```diff
@@ -363,14 +363,28 @@
...
+      ansible.builtin.find:
+        paths: /etc/logrotate.d/
+        patterns: "skynet-webportal-*"
+      register: existing_logrotate
+
+    - name: Remove all existing skynet-webportal logrotate configs
+      file:
+        state: absent
+        path: "{{ item.path }}"
+      with_items: "{{ existing_logrotate.files }}"
```
firyx, line 376: `with_items` is deprecated.

```suggestion
      loop: "{{ existing_logrotate.files }}"
```

firyx (changes requested):

firyx (approved):

